PR #25, opened 2020-05-12
Add simple to-do app to examples repo

lsgunnlsgunn:
**Overview**
The simple to-do checklist app is an example that is used in the documentation. Might as well be included here, too.

**Requirements**
n/a

**Considered Solutions**
n/a

**Recommended Solution**
The more examples, the merrier!


motoko/simple_to_do/src/simple_to_do/public/index.js
```diff
@@ -0,0 +1,5 @@
+import simple_to_do from 'ic:canisters/simple_to_do';
+
+simple_to_do.addTodo(window.prompt("Enter your task")).then(Todos => {
+  window.alert(Todos);
+});
```
stanleygjones, line 5: I'd like to revisit this example at some point to use an off-the-shelf front-end from [TodoMVC](http://todomvc.com).

motoko/simple_to_do/src/simple_to_do/utils.mo
```diff
@@ -0,0 +1,53 @@
...
+// Clear to-do item utility
+func clear(todos: [ToDo]) : [ToDo] {
+	var updated : [ToDo] = [];
+	for (todo : ToDo in todos.vals()) {
+		if (not todo.completed) {
+			updated := Array.append<ToDo>(updated, [todo]);
+		};
+	};
+	updated
+};
```
stanleygjones, line 53: In retrospect, this should probably be a filter, something like:
```
func clear(todos: [ToDo]) : [ToDo] {
  Array.filter<ToDo>(func (todo: ToDo) : Bool { not todo.completed }, todos);
}
```

motoko/simple_to_do/src/simple_to_do/utils.mo
```diff
@@ -0,0 +1,53 @@
...
+// Add to-do item utility
+func add(todos : [ToDo], desc : Text, nextId : Nat) : [ToDo] {
+	let todo : ToDo = {
+	id = nextId;
+	description = desc;
+	completed = false;
+};
+	Array.append<ToDo>(todos, [todo])
+};
+
```
stanleygjones, line 19: The indentation looks a little funky here, I think it should be:
```
func add(todos : [ToDo], desc : Text, nextId : Nat) : [ToDo] {
  let todo : ToDo = {
    id = nextId;
    description = desc;
    completed = false;
  };
  Array.append<ToDo>(todos, [todo])
};
```

motoko/simple_to_do/src/simple_to_do/public/index.js
lsgunnlsgunn: Yeah, something that works would be cool. I tried to fake something up to replace the greet/greeting frontend piece but didn't want to add a full React type thing so I punted.

PR #29, opened 2020-05-28
improvement(frontend): add React boilerplate example

andrewwylde:
**Overview**
This helps developers who are unfamiliar with the front-end ecosystem to get up and running using React, one of the more popular frameworks/libraries considered for front-end usage.


frontend/dfx_react/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+- [Quick Start](https://sdk.dfinity.org/developers-guide/quickstart.html)
+- [Developer's Guide](https://sdk.dfinity.org/developers-guide)
+- [Language Reference](https://sdk.dfinity.org/language-guide)
+
+## Important Note
+
+Before you run `dfx build`, you must update the `@dfinity/agent` property in the package.json.
+
+To get the location of your `@dfinity/agent`, run `echo $HOME/.cache/dfinity/versions/<your-dfx-version>/js-user-library`, which should print something like this:
```
hansl, line 15: Is this internal documentation? We plan to publish @dfinity/agent publicly with the next release.

hansl: Also, `echo $(dfx cache show)/js-user-library` will do the same.

frontend/dfx_react/src/dfx_react/public/index.jsx
```diff
@@ -0,0 +1,42 @@
+import "regenerator-runtime/runtime.js";
+
+import reactCanister from 'ic:canisters/dfx_react';
+
+import React, { useState } from 'react';
+import ReactDOM from 'react-dom';
+
+async function asyncUpdateCount(val, callback) {
```
hansl, line 8: I noticed this in both examples. Is there a reason to use `val` as a string and use a switch/case?

See https://github.com/dfinity/examples/pull/30/files#r433553269.

codelemur:
Same fixes from #30

PR #39, opened 2020-08-23
Add makefile example

gobengo:
**Overview**

Demonstrate a reasonable way of using gnu `make` with a dfx project.

Other projects already have Makefiles, but they are quite [basic](https://github.com/dfinity/examples/blob/master/motoko/echo/Makefile). 

**Requirements**

* Add a new example called 'makefile'
* In this project, the Makefile should be more advanced.
* I should be able to run `make call method={actorMethod} [arg]` to create the canister, build the src code, install to cansiter, and then call the canister.
* As much as possible, try to cache artifacts between calls using make dependencies

**Considered Solutions**

Just this one.

**Recommended Solution**

Fairly straightforward use of `make`, though I'm not an expert.

**Considerations**

It is a new example to 'maintain' over time.
But it should be useful as a place to go copypasta a good starting point Makefile for any new dfx project.
* I tested this by pasting into one of the 'hello' tutorial projects, and:
  ```
  ▶ make call method=greet ben
  dfx canister call hello greet "ben"
  ("Hello (v17), ben!")
  ```
If it's useful enough, maybe it even makes it into the core project scaffold's Makefile (output of `dfx new`).

### DIFF
--- a/motoko/makefile/Makefile
+++ b/motoko/makefile/Makefile
@@ -1,0 +1,59 @@
+.PHONY: default canister build install call
+
+src := $(wildcard src/*/*.mo) $(wildcard src/*.mo)
+
+c ?= $(shell basename $(CURDIR))
+canister ?= $(c)
+
+m ?= ''
+method ?= $(m)
+
+a ?= ''
+args ?= $(a)
+
+install_mode ?= reinstall
+
+# commands
+
+default:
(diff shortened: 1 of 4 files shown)


dfx-json:
hi @gobengo. thank you for contribution. as this PR is quite outdated, I would like to close it. if you would still like to include this example, I encourage you to create a new PR. thank you for your understanding.

Issue #40, opened 2020-08-28
Failed to run hello-world sample


dimas-prawira:
Hi, I have download SDK based on instructions and testing it with command

`$dfx --version
dfx 0.6.4
`

I clone examples motoko, from github and start to run hello-world based on instructions 

1. run `$dfx start`
    `sudo dfx start
⠦ Checking replica!
Aug 28 14:52:16.142 INFO ic-starter. Configuration: ValidatedConfig { node_id: 100, replica_path: Some("/root/.cache/dfinity/versions/0.6.4/replica"), subnet_id: 0, cargo_bin: "cargo", cargo_opts: "", state_dir: "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replicated_state", http_port_config: WritePortTo("/home/dimas/motoko/examples/motoko/hello-world/.dfx/client-configuration/client-1.port"), http_require_valid_signatures: true, artifact_pool_dir: "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replicated_state/node-100/ic_consensus_pool", crypto_root: "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replicated_state/node-100/crypto", state_manager_root: "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replicated_state/node-100/state", registry_file: "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replicated_state/node-100/registry.proto", state_dir_holder: None }, Application: starter
Aug 28 14:52:16.142 INFO Executing "/root/.cache/dfinity/versions/0.6.4/replica" "/home/dimas/motoko/examples/motoko/hello-world/.dfx/state/replic⠒ Checking replica!
Aug 28 14:52:16.525 WARN s:0/n:[100, 0, 0, 0, 0, 0, 0, 0, 253, 1]/ic_p2p/p2p P2P::process_messages(): processing msg_types = 1
Aug 28 14:52:16.525 ERRO s:0/n:[100, 0, 0, 0, 0, 0, 0, 0, 253, 1]/ic_messaging/xnet_endpoint No XNet configuration for node 5wjwg-gteaa-aaaaa-aaaap-2ai. This is an error in production, but may be ignored in single-subnet test deployments.
Aug 28 14:52:16.525 WARN s:0/n:[100, 0, 0, 0, 0, 0, 0, 0, 253, 1]/ic_p2p/p2p P2P::process_messages(): processing msg_types = 4
[...]

PR #41, opened 2020-08-28
Fix hello-world source and README to work

lsgunnlsgunn:
**Overview**
As reported in https://github.com/dfinity/examples/issues/40, this example and instructions are out-of-date.
We are sending people to this repo so we should make sure the examples work.

### DIFF
--- a/motoko/hello-world/README.md
+++ b/motoko/hello-world/README.md
@@ -11,22 +11,27 @@ Before building the example application, verify the following:
 
 ### Demo
 
-Start a local internet computer.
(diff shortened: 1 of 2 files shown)


crusso (approved):
LGTM - thanks for doing this - I wonder if the other samples are broken too.

FTR, Did you actually verify these steps work?

lsgunnlsgunn:
I did indeed 😀

lsgunnlsgunn:
I did indeed 😀

PR #42, opened 2020-08-28
Add the register step

lsgunnlsgunn:
**Overview**
Looks like all of the examples need to be updated.

### DIFF
--- a/motoko/simple_to_do/README.md
+++ b/motoko/simple_to_do/README.md
@@ -21,13 +21,19 @@ Before building the example application, verify the following:
 
 ## Demo
 
-1. Start a local internet computer replica by running the following command:
+1. Start the Internet Computer network locally by running the following command:
 
     ```bash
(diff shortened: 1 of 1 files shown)


crusso (approved):
LGTM

matthewhammer:
LGTM as well.

FWIW, I find it slightly displeasing that we say `dfx canister create` and refer to this as "registering" the canister.

Seems like we should explain how the verbs "create" and "register" relate, somehow, somewhere.  (Perhaps start here, with me? Honest question.)

lsgunnlsgunn:
I have been lobbying for "create" to be changed to "register" in both the spec and DFX. Register (and deploy) are used to describe the "create" (and "install") functionality so it seems reasonable to use that terminology directly. The create might have made its way into a PR for the public spec and we've been talking about it in our SDK design/arch meetings.

HTH

PR #43, opened 2020-09-01
Add a simple system time example

lsgunnlsgunn:
**Overview**
Add a very simple example for using the Time module from motoko-base.


chenyan-dfinity:
Ah, I see. The Motoko base in dfx is not in sync with `master`. Maybe we need to wait the example for the next release. The current Time module is not the interface we want to expose.

lsgunnlsgunn:
Closing this for now since there's a disconnect between the time exposed in dfx and the current motoko-base

PR #47, opened 2020-09-16
Update Simple_to_do for latest Base

matthewhammer:
(Follow latest argument order style: We decided to place the order of arguments that are themselves functions at the end.)


lsgunnlsgunn (approved):
Awesome. Thanks @matthewhammer !

PR #49, opened 2020-10-02
Updating reverse.c example for latest tool chain (clang-11) and dfx 0.6.10

skilesare:
**Overview**

The old example no longer works.

**Recommended Solution**

Updating the dfx.json format and adding a reverse.did file.  I also provide a script to do this in a container.

**Considerations**

None.

### DIFF
--- a/c/reverse/README.adoc
+++ b/c/reverse/README.adoc
@@ -40,7 +40,7 @@ If you are not using the docker container you need to install Clang:
 
 RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
 
-and set up a couple of aliases. navigate to the examples/c/reverse directory. Set up the following aliases:
+and set up a couple of aliases. Navigate to the `examples/c/reverse` directory. Set up the following aliases:
 
   $ alias wld="wasm-ld-11 --no-entry --export-dynamic --allow-undefined"
   $ alias wcc="clang-11 --target=wasm32 -c -O3"
--- a/c/reverse/README.adoc
+++ b/c/reverse/README.adoc
@@ -42,6 +42,7 @@ RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
 
 and set up a couple of aliases. Navigate to the `examples/c/reverse` directory. Set up the following aliases:
 
+``` sh
   $ alias wld="wasm-ld-11 --no-entry --export-dynamic --allow-undefined"
   $ alias wcc="clang-11 --target=wasm32 -c -O3"
   
--- a/c/reverse/README.adoc
+++ b/c/reverse/README.adoc
@@ -3,11 +3,49 @@
 DFINITY uses standard wasm binaries, so we may use standard tools such as Clang
 to build apps.
 
-First, install Clang and set up a couple of aliases. For example, on Debian:
+# Running in Docker
 
-  $ sudo apt-get install clang lld gcc-multilib
-  $ alias wld="wasm-ld-8 --no-entry --export-dynamic --allow-undefined"
-  $ alias wcc="clang --target=wasm32 -c -O3"
+To create a docker container with all the libraries required and dfx follow the below instructions.  If you already have a debian environment you'd like to use, skip to the following section.
+
+ Create a docker file with the below text:
+ 
+  FROM debian
+  RUN apt-get update
+  RUN apt-get install -y curl gnupg xz-utils nodejs npm wget git
+  RUN apt install -y lsb-release wget software-properties-common
+  RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
+  CMD ["/bin/bash"]
+  WORKDIR /usr/src
+ 
+Create the new docker container by running the following command 
+
+  $ docker build --tag debian:dfx .
+
+Start the instance with 
+
+  $ docker run -ti --name dfx debian:dfx
+  
+This will launch you into your instance. Now we need to pull the code by running:
+
+  $ cd usr/src
+  $ git clone https://github.com/dfinity/examples.git
+  
+Install the dfinity sdk using
+
+  $ sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+  
+# Compiling the C code  
+
+If you are not using the docker container you need to nstall Clang:
+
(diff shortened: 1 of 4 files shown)


c/reverse/README.adoc
```diff
@@ -3,11 +3,49 @@
...
+Start the instance with 
+
+  $ docker run -ti --name dfx debian:dfx
+  
+This will launch you into your instance. Now we need to pull the code by running:
+
+  $ cd usr/src
+  $ git clone https://github.com/dfinity/examples.git
+  
+Install the dfinity sdk using
```
ggreif, line 33:
```suggestion
Install the DFINITY sdk using
```

c/reverse/README.adoc
```diff
@@ -3,11 +3,49 @@
...
+
+  $ sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+  
+# Compiling the C code  
+
+If you are not using the docker container you need to nstall Clang:
```
ggreif, line 39:
```suggestion
If you are not using the docker container you need to install Clang:
```

c/reverse/README.adoc
```diff
@@ -3,11 +3,49 @@
...
+
+RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
+
+and set up a couple of aliases. navigate to the examples/c/reverse directory. Set up the following aliases:
```
ggreif, line 43:
```suggestion
and set up a couple of aliases. Navigate to the examples/c/reverse directory. Set up the following aliases:
```

c/reverse/README.adoc
```diff
@@ -3,11 +3,49 @@
...
+
+  $ alias wld="wasm-ld-11 --no-entry --export-dynamic --allow-undefined"
+  $ alias wcc="clang-11 --target=wasm32 -c -O3"
```
ggreif, line 46:
```suggestion
``` sh
  $ alias wld="wasm-ld-11 --no-entry --export-dynamic --allow-undefined"
  $ alias wcc="clang-11 --target=wasm32 -c -O3"
```
```

File: c/reverse/README.adoc [hunk 5]
```diff
@@ -17,22 +55,26 @@ wasm binary with:
...
-  $ mkdir build
-  $ mkdir build/reverse
-  $ cp reverse.wasm build/reverse/
-  $ head -c 8 /dev/urandom > build/reverse/_canister.id
-
-The script `build.sh` performs the above steps.
+  $ dfx canister create reverse
+  $ dfx build
+  
+You now need a seperate terminal (docker exec -ti dfx bash) to start the dfx server. Make sure you are in the reverse directory and run
```
ggreif, line 63:
```suggestion
You now need a separate terminal (`docker exec -ti dfx bash`) to start the local replica. Make sure you are in the reverse directory and run
```

ggreif (approved):
Thanks for doing this! Minor nits in the comments.

c/reverse/reverse.c
```diff
@@ -1,14 +1,4 @@
...
-  char*t = s+len-1,c;
-  while(s<t) {
-    c = *t;
-    *t = *s;
-    *s = c;
-    s++;
-    t--;
-  }
-}
+#include "duktape.h"
```
chenyan-dfinity, line 1: Is this file missing?

c/reverse/README.adoc
```diff
@@ -3,11 +3,49 @@
...
-  $ alias wld="wasm-ld-8 --no-entry --export-dynamic --allow-undefined"
-  $ alias wcc="clang --target=wasm32 -c -O3"
+To create a docker container with all the libraries required and dfx follow the below instructions.  If you already have a debian environment you'd like to use, skip to the following section.
+
+ Create a docker file with the below text:
+ 
+  FROM debian
+  RUN apt-get update
+  RUN apt-get install -y curl gnupg xz-utils nodejs npm wget git
+  RUN apt install -y lsb-release wget software-properties-common
```
chenyan-dfinity, line 15: Are these required by llvm? dfx should be standalone and not depend on other packages. Pretty sure we don't need `nodejs` and `npm`, as there is no front end here.

skilesare: I had to install them to get some things to work.  Basically dfx build was failing when it was trying to build some things.  I can't find the error now, but it was right after the output of something like 'setting up the environment'.  My understanding was that node was required for dfx. Maybe that has changed in the latest release?  I'll try to replicate with out those packages and see.

c/reverse/reverse.c
skilesare: ahh...sorry... I didn't realize that it would continue to track master.  Let me revert and move this to a branch.

c/reverse/README.adoc
chenyan-dfinity: With the new dfx.json file, `dfx build` is simply copying wasm files around. It won't do any actual build.

c/reverse/reverse.c
chenyan-dfinity: Why do you need to rely on the `duktape` library? Do you get errors with the original C code?

skilesare: Ok...that file should be back to the original now.

skilesare: Duktape is an experiment I'm trying to do to get a javascript parser running inside of a canister. I've moved it to a fork for messing around. It was just easier to use the already set up reverse calls and wasm imports in that example to try to get it to work because I don't really understand the API and it isn't documented anywhere.(I haven't gotten it to work....any help would be awesome.)

c/reverse/README.adoc
skilesare: Your right! You don't need node for this example so I removed it. If you want to play around with any other dfx commands like new you do need it or you get:

Node could not be found. Skipping installing the frontend example code.

I've removed them for the time being.

c/reverse/reverse.c
skilesare: If you have some tips I'm posting over here:  https://forum.dfinity.org/t/help-with-candid-file-construction/1160/7  I'd also love to get back on the dfinity slack.  I think my previous access was revoked.

enzoh:
If you're looking for a better example that demonstrates how to build a canister for the Internet Computer in C, then I would recommend checking out https://github.com/enzoh/counter.

dfx-json:

Issue #51, opened 2020-10-02
pub_sub example won't work.


miguelemosreverte:
Hello there, I wanted to create a benchmark of this project and landed on the examples which looked promising.

However, this happened. It fails to build!

**$ dfx canister create sub**
Creating canister "sub"...
"sub" canister created with canister id: "do2cr-xieaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q"
**$ dfx canister create pub**
Creating canister "pub"...
"pub" canister created with canister id: "6ttj4-pafaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q"
**$ sh test.sh** 
Building canisters...
Build failed. Reason:
  Build step failed for canister 6ttj4-pafaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q with error: Build failed. Reason:
  Command "/home/miguelemosreverte/.cache/dfinity/versions/0.6.10/moc" "/home/miguelemosreverte/Desktop/motoko/examples/motoko/pubsub/src/pub/main.mo" "-o" "/home/miguelemosreverte/Desktop/motoko/examples/motoko/pubsub/.dfx/local/canisters/pub/pub.did" "--idl" "--actor-idl" "/home/miguelemosreverte/Desktop/motoko/examples/motoko/pubsub/.dfx/local/canisters/idl/" "--actor-alias" "pub" "6ttj4-pafaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q" "--actor-alias" "sub" "do2cr-xieaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q" "--package" "base" "/home/miguelemosreverte/.cache/dfinity/versions/0.6.10/base"
 returned an error:
/home/miguelemosreverte/Desktop/motoko/examples/motoko/pubsub/src/pub/main.mo:8.1-22.2: type error, an actor or actor class must be the only non-imported declaration in a program
  (This is a limitation of the current version.)

Installing code for canister pub, with canister_id 6ttj4-pafaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-q
An error occured:
Io(
    Os {
        code: 2,
        kind: NotFound,
        message: "No such file or directory",
    },
)
cannot find method type, dfx will send message with inferred type
[...]

PR #52, opened 2020-10-05
fix: update all motoko examples to use v0.6.10 version of `dfx`

codelemur:
Major breaking change in Motoko meant that all typings need to exist _inside_ of the actors, and I figured I'd pin each example to a version to make sure folks know they're running the right `dfx`. Verified each of these locally to make sure things built and UIs were updated to the current best practice pattern (i.e. asset canisters).


motoko/simple_to_do/test.sh
```diff
@@ -0,0 +1,16 @@
...
+dfx canister call simple_to_do addTodo '("Build the project")'
+dfx canister call simple_to_do addTodo '("Deploy the project")'
+
+# view todos
+dfx canister call simple_to_do showTodos
+
+# complete and observe change
+dfx canister call simple_to_do completeTodo '(1)'
+sleep 2
+dfx canister call simple_to_do showTodos
```
crusso, line 16:
```suggestion
dfx canister call simple_to_do showTodos

```

crusso (approved):
LGTM, thanks for doing this

miguelemosreverte:
this fixes Issue #51

PR #53, opened 2020-10-06
Add a badge for compatibility

lsgunnlsgunn:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


hansl (approved):
Great idea!

PR #76, opened 2021-03-16
Example: Usernames and Passwords

ghost:
We sometimes talk about how "you don't need usernames and passwords" for the Internet Computer, but that doesn't mean you can't use them... and in some ways they make it easier to juggle multiple devices.


motoko/passwords/README.md
```diff
@@ -0,0 +1,78 @@
...
+1. Create a new account.
+
+   ```text
+   dfx canister call passwords signup 'record { username = "Alice"; password = "p4ssw0rd" }'
+   ```
+
+1. Authenticate your command line identity.
+
+   ```text
+   dfx canister call passwords login 'record { username = "Alice"; password = "p4ssw0rd" };
```
gobengo, line 55: It might be a good idea to put a big warning here saying "DO NOT USE A PASSWORD YOU USE ON OTHER SITES". (It's my understanding we currently really make no promises about this value staying confidential from people who shouldn't have it. Dom might have a plan, but IMO he hasn't made a convincing argument that it's safe to tell the IC real secrets).

Even better, dont' take the risk at all. Do the password hashing (ideally [salted](https://en.wikipedia.org/wiki/Salt_(cryptography))... if we had a `serviceId` that might be a good salt) here in the shell.

Instead of storing `password`, store something like `password_hash` or, better, can even support a few password hash schemes, so like `password_hash_bcrypt` https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms

A naive hashing scheme would just be md5, and if it were, this line could be replaced with:

```
password_hash_md5=$(echo-my-password-to-stdout | md5)
dfx canister call passwords login 'record { username = "Alice"; password_hash_md5 = "$password_hash_md5" };
```

Now my password never was sent to IC.

There are probably more things in OWASP worth checking this against. I only looked it up to check if `bcrypt` is still a good one. https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#introduction

motoko/passwords/dfx.json
```diff
@@ -0,0 +1,34 @@
...
+      ],
+      "type": "assets"
+    }
+  },
+  "defaults": {
+    "build": {
+      "packtool": ""
+    }
+  },
+  "dfx": "0.6.24",
```
gobengo, line 26: does it work with `dfx@latest`? 0.6.25

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+  var accounts : [Account] = [];
+  var sessions : [Session] = [];
+  var next_account_id : Nat = 0;
+
+  // Public Functions
+
+  public func signup (credentials : Credentials) {
+    // Generate a random salt and hash it together with the password
+    let _salt = random_salt();
+    let _hash = sha256(credentials.password # _salt);
```
gobengo, line 39: `bcrypt` (et al) may get less complaints than `sha256`. My understanding is that `sha256` is too fast to calculate, so too succeptible to brute-force attacks (if you can access the 'secret' password hash, then just guess all the passwords and hash each until you find a hash that matches)

https://cwe.mitre.org/data/definitions/916.html
https://dusted.codes/sha-256-is-not-a-secure-password-hashing-algorithm

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+
+    // If the hashed passwords match then create a session
+    if (sha256(credentials.password # account.salt) == account.hash) {
+      // The session associates this caller's ID with the account
+      let session : Session = {
+        principal = caller;
+        account_id = account.id;
+      };
+      sessions := Array.append(sessions, [session]);
+    };
```
gobengo, line 66: note: subsequent lies added later will execute even if password doesn't match.

I get scared and prefer writing:
```
    if (sha256(credentials.password # account.salt) != account.hash) {
      return NotAuthorizedResponse()
    }
```

Slightly related, as someone calling this method and getting a return value, how do I know if the request I just sent to log in was valid or not? (maybe intentionally left opaque?)

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  public shared ({ caller }) func logout () {
+    // Remove the session for this caller's ID
+    sessions := Array.filter(sessions, omitCaller(caller));
```
gobengo, line 71: cycle cost will be O(n), with HashMap closer to O(1), and also I think it's fine to keep using arrays. (may be worthwhile to educate on the cost-downsides in a comment or something)

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+    let session = Option.unwrap(Array.find(sessions, findByCaller(caller)));
+    let account = Option.unwrap(Array.find(accounts, findByAccountId(session.account_id)));
+    return account.username;
+  };
+
+
+  // "Crypto" Library
+
+  func sha256 (raw : Text) : Text {
+    // WARNING: This is a terrible hashing algorithm
```
gobengo, line 85: Nice! Maybe help dev out with link to a good source of other ones? https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+    raw
+  };
+
+  func random_salt () : Text {
+    // WARNING: This is a terrible random salt generator
+    "salt"
```
gobengo, line 91: If it is easy to get a random int (idk), a slight upgrade may be js equiv of:

```
return 'abcdefghijklmnopqrstuvwxyz'[Math.floor(Math.random() * 26)]
```

if not, the comment tells me not to use it as-is. good one

motoko/passwords/src/passwords/main.mo
```diff
@@ -0,0 +1,121 @@
...
+  // Data Structures
+
+  // Arrays a sub-optimal choice here but they're easy to understand
+  var accounts : [Account] = [];
+  var sessions : [Session] = [];
+  var next_account_id : Nat = 0;
+
+  // Public Functions
+
+  public func signup (credentials : Credentials) {
```
gobengo, line 36: I think this might have some indeterminate behavior if different users both try to `signup` with 'identical credentials', i.e. a Credentials with identical `username` and `password.

i.e. I think there are no checks to make sure that the username/password aren't already taken by someone else, and allowing access to that someone else's account. (correct me if I'm wrong. Hard for me to tell if this should happen in `signup` or `login`)

motoko/passwords/src/passwords_assets/public/index.js
```diff
@@ -0,0 +1,35 @@
+import passwords from "ic:canisters/passwords";
```
gobengo, line 1: Could be nice if this would do `passwords.signup` as well, esp to test what happens if I try to signup as the same username/password as someone else.

motoko/passwords/src/passwords/main.mo
gobengo: I like how you called `random_salt()` a good name even with a stubbed implementation.
Could be illustrative to also have a `hashPassword(password, _salt)` function (that for now is just stubbed to call out to 'bad' `sha256). Purely editorial decision by you tho.

gobengo (changes requested):
Requested changes:
* Help me understand what might happen with duplicate/identical Credentials (i.e. same user/pass for two Accounts) https://github.com/dfinity/examples/pull/76#discussion_r595612656
* IMO we should either not encourage end-users to submit plaintext passwords in ICP requests..., or at the very least contextualize the risks of this. https://github.com/dfinity/examples/pull/76/files/7f35f73d02783b34ffe59bb7ed123a8355b8615c#r595598060


Nice to have:
* Include a comment that using Arrays and not HashMaps has not only a simplicity tradeoff, but a cycle-cost one (afaict) https://github.com/dfinity/examples/pull/76/files/7f35f73d02783b34ffe59bb7ed123a8355b8615c#r595608392
* Is motoko-side hashing even safe to illustrate at all? vs client-side routing (e.g. end-user only sends password_hash_bcrypt over ICP messages)
* If it works with `dfx@latest` -> `0.6.25` (rn), might as well check it in that way. Nice-to-have because it's easy to check later. https://github.com/dfinity/examples/pull/76/files/7f35f73d02783b34ffe59bb7ed123a8355b8615c#r595601196
* Link in comment to better hashing functions than `sha256` https://github.com/dfinity/examples/pull/76/files/7f35f73d02783b34ffe59bb7ed123a8355b8615c#r595609927
* Little more education on *why* this is a bad random_salt https://github.com/dfinity/examples/pull/76/files/7f35f73d02783b34ffe59bb7ed123a8355b8615c#r595610502

stanleygjones:
@gobengo Thanks for the thorough review:

For the requested changes...

- Fixed! Good suggestion, I totally overlooked that case.
- I added a warning about sending passwords in the clear, despite being over HTTPS.

For the other comments...

- HashMaps or even just Maps would definitely be better for Sessions, which we only ever lookup by `caller` but we lookup Accounts by both `username` and `account_id` so I don't think we'd gain much from a HashMap there.
- Let me think a little more about client-side hashing. I think the example gets my points across for now.
- Oh, 0.6.24 was the latest on my machine. I can probably bump it. I'll check.
- I swapped in `bcrypt`. I was only using `sha256` because I was hoping to use the existing Motoko library for it.
- I made an attempt to actually generate a random salt. It was unfortunately a rather complex affair. You have to grab randomness from the management canister 32 bytes at a time, which is asynchronous. I think a better solution would be to use the randomness to seed a local pseudo-random generator that would be available synchronously, but that started to inflate the example too much.

stanleygjones:
Bumped `dfx` to 0.6.25. Smooth sailing.

sa-github-api:
Dear @ghost,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/437) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

dfx-json:

PR #77, opened 2021-04-11
fix 🐛 no such file or dir (os error 2)

heldrida:
**Overview**
When running the example shell scripts an error is thrown "No such file or directory (os error 2)".
These changes provide a fix.

**Requirements**
Checkout the branch and run ./build.sh && ./start.sh

**Considered Solutions**
n/a

**Recommended Solution**
On dfx 0.6.26 the directory is expected to be `.dfx/local/canisters/reverse`,
for that reason a mkdir command creates the desired build directory.

The script now also skips build if the directory already already exists,
preceding the call to the canister reverse service.

On sucessful build, the wasm file is now moved instead of copied,
and artifacts are cleared from root.

**Considerations**
n/a


dfx-json:

heldrida:
Sure, 3 years no one cared to have a look :)

PR #79, opened 2021-04-22
Game of Life and upgrades example; Upgrade all to  0.7.0-beta.3

crusso:
**Overview**

Add Game Of Life Upgrades demo; bump dfx to version 0.7.0-beta.3; adjust existing examples to Motoko changes (WordX -> NatX etc.)

@hansl  - I had to disable an asset canister test in the phone-book example. Can you suggest a suitable replacement for that test? Here's the context:

motoko/phone-book/Makefile:

```
test: install
	dfx canister call phone_book insert '("Police", record { desc = "Call if you have an emergency."; phone = "911" })'
	dfx canister call phone_book lookup '("Police")' \
		| grep '911' && echo 'PASS'
	# This test needs to be rewritten for the new HTTP based architecture
	# dfx canister call www retrieve '("index.js")' \
	#	| grep 'Phone Book' && echo 'PASS'
```

TODO:
- [ ] adapt disabled test (see above)
- [x] check test with old-style front-ends actually work in browser
- [x] investigate phone-book/superhero npm build errors (that strangely don't fail CI)


taylorham (approved):
🙌

PR #81, opened 2021-04-29
Simplified, extended and tested hello_cycles example

crusso:
This continues and completes the example @lsgunnlsgunn started to illustrate transferring cycles between canisters and wallets.

Doesn't add the access control checks to prevent people stealing cycles but at least warns of the possibility.

Not great, but not terrible either.

(Writing tests in bash and make is painful. Spent way too much time on this.)


lsgunnlsgunn (approved):
Thanks, Claudio!

PR #95, opened 2021-05-25
Claudio/classes

crusso:
An example illustrating actor classes, with cycles.

### DIFF
--- a/motoko/classes/Makefile
+++ b/motoko/classes/Makefile
@@ -1,0 +1,31 @@
+.PHONY: all
+all: build
+
+.PHONY: build
+.SILENT: build
+build:
+	dfx canister create counter
+	dfx build
+
+.PHONY: install
+.SILENT: install
+install: build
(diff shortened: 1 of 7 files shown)


motoko/classes/Makefile
```diff
@@ -0,0 +1,29 @@
...
+.PHONY: test
+.SILENT: test
+test: install
+	dfx canister call Test run '()' \
+		| grep '()' && echo 'PASS'
+
+.PHONY: clean
+.SILENT: clean
+clean:
+	rm -fr .dfx
```
ggreif, line 29: I usually write `@ rm ...` and then the `.SILENT` is  not needed. For one-liners.

ggreif (approved):

motoko/classes/Makefile
crusso: thanks - I'm just cargo-cult coding here so will keep to same style.

PR #101, opened 2021-06-06
Add dist/ to .gitignore

johnkruger37:
This directory is produced by Node.JS during `dfx deploy`, and should be ignored.

It is on the `.gitignore` produced by `dfx new` as well ([see e.g. here](https://github.com/johnkruger37/learn-internetcomputer/blob/main/.gitignore)).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .dfx/
 build/
+dist/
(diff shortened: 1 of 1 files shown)


dfx-json:

PR #103, opened 2021-06-20
Added clarity as to why...

cPlayIt:
Is this correct?

**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/motoko/pub-sub/README.md
+++ b/motoko/pub-sub/README.md
@@ -7,7 +7,7 @@ This sample project demonstrates how functions may be passed as arguments of int
 
 ## Overview
 
-A common problem in both distributed and decentralized systems is keeping separate services (or canisters) synchronized with one another. While there are many potential solutions to this problem, a popular one is the Publisher/Subscriber pattern or "PubSub". PubSub is an especially valuable pattern on the Internet Computer as its primary drawback, message delivery failures, does not apply.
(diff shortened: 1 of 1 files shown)


motoko/pub-sub/README.md
```diff
@@ -7,7 +7,7 @@ This sample project demonstrates how functions may be passed as arguments of int
 
 ## Overview
 
-A common problem in both distributed and decentralized systems is keeping separate services (or canisters) synchronized with one another. While there are many potential solutions to this problem, a popular one is the Publisher/Subscriber pattern or "PubSub". PubSub is an especially valuable pattern on the Internet Computer as its primary drawback, message delivery failures, does not apply.
+A common problem in both distributed and decentralized systems is keeping separate services (or canisters) synchronized with one another. While there are many potential solutions to this problem, a popular one is the Publisher/Subscriber pattern or "PubSub". PubSub is an especially valuable pattern on the Internet Computer as its primary drawback, the potential for lock-ups due to message delivery failures, is typically a non-issue due loose coupling between the publisher and subscriber in a way that each can continue to operate normally when a receiving side [or the network between] is not.
```
crusso, line 10:
```suggestion
A common problem in both distributed and decentralized systems is keeping separate services (or canisters) synchronized with one another. While there are many potential solutions to this problem, a popular one is the Publisher/Subscriber pattern or "PubSub". PubSub is an especially valuable pattern on the Internet Computer as its primary drawback, the potential for lock-ups due to message delivery failures, is typically a non-issue due to loose coupling between the publisher and subscriber in a way that each can continue to operate normally when a receiving side [or the network between] is not.
```

crusso: Actually, I'm not even sure what the original was trying to see. See here https://forum.dfinity.org/t/pub-sub-example-message-delivery-failures-confusion/5534

dfx-json:
thank you for your contribution. as this PR is quite outdated, I would like to close it. if you would still like to submit your changes, please open a new PR. thank you for your understanding!

PR #108, opened 2021-06-23
update to 0.7.2

crusso:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


krpeacock (approved):
The change to the test cases is appropriate after the default balance changed in 0.7.2

ggreif (approved):

PR #115, opened 2021-08-10
More idiomatic Wasm style

rossberg:
With this being an example for public consumption, I couldn't resist. :)


ggreif:
Not that I expect to ever write Wasm by hand, but the diff was very helpful, thanks!

ggreif (approved):

chenyan-dfinity (approved):
LGTM :)

PR #118, opened 2021-10-04
index the Motoko examples, with categories

matthewhammer:
Adding an index file so that readers can get an overview of the Motoko examples without descending into and reading each one.


matthewhammer:
@crusso @chenyan-dfinity I have a few questions.

I'm still getting caught up with the start of the art of "front end canister development" using dfx.  As I read these examples, I found myself making this mental categorization of them.  I'm adding this index file here in case it's helpful generally, but also, just to double-check that I've classified the examples correctly myself.  That's the primary purpose of this PR.

In particular, I'm still trying to appreciate the role of `react` and `react-dom` in the `package.json` files of these projects.

(Keep in mind that I know basically nothing about how frontend development is supposed to work here, so my questions are basic.)

In at least three cases, the role of React is fairly clear to me as a novice: It defines the class that defines the javascript behavior for the application, via a line like `class Life extends React.Component { ...` in a file like `index.jsx` in the assets for the project.

However, that does not seem to be the case for `random_maze`, which seems more minimal, and thus worth imitating for further examples that are supposed to be minimal, like certified variables, my current task.  So, I plan to dig into that more, to understand how it builds and deploys, with the hope of doing something similar for the forthcoming certified var example.

(Can you confirm that I'm saying accurate things here, and in the README file?)

chenyan-dfinity:
It's more or less correct. But I see both `random_maze` and other examples with frontend as full-stack examples. The use of a particular library/framework (in this case "react") doesn't make them different from other apps.

React is there to make some incremental DOM updates "easier", but it's not a requirement to build the frontend on the IC. In this sense, you can say `random_maze` and the default example from `dfx new` are minimal examples in the sense that you are just writing plain JavaScript, and not some DSL (React/Vue/Elm are all DSLs) built on top of JS.

Of course, you can go even further/minimal by removing the agent-js package with a single file (see Paul's example: https://github.com/ninegua/ic-qr-scanner/blob/main/src/bare-agent.js). But that can make user code more difficult to write in some cases. So there is a balance for a minimal setup :)

matthewhammer:
> Looks useful, thanks. Did you consider making those headings links?

Good idea -- Done!

matthewhammer:
@chenyan-dfinity Did you want me to adjust the organization of the examples with a frontend, or how to term them?

Perhaps I will merge this now, and we can revisit again?  I agree that the terminology is not yet ideal.

PR #119, opened 2021-10-04
adapting frontend assets for cert_var from random_maze

matthewhammer:
Shows a minimal demonstration of an IC-specific system feature: certified data, available to fast queries.

Consists of a Motoko canister that holds a single certified variable, as a 32-bit number, and an asset canister whose (client-side) code queries and certifies this number and its certificate.

### DIFF
--- a/motoko/cert-var/README.md
+++ b/motoko/cert-var/README.md
@@ -1,0 +1,54 @@
+# Certified Variable
+
+![Compatibility](https://img.shields.io/badge/compatibility-0.7.0-blue)
+[![Build Status](https://github.com/dfinity/examples/workflows/motoko-cert_var-example/badge.svg)](https://github.com/dfinity/examples/actions?query=workflow%3Amotoko-cert_var-example)
+
+The example shows a cryptographically certified variable.
+
+
+
+## Prerequisites
+
+Verify the following before running this demo:
+
+*  You have downloaded and installed the [DFINITY Canister
+   SDK](https://sdk.dfinity.org).
+
+*  You have stopped any Internet Computer or other network process that would
+   create a port conflict on 8000.
+
+## Demo
+
+1. Start a local internet computer.
+
+   ```text
+   dfx start
+   ```
+
+1. Open a new terminal window.
+
+2. Install the front-end dependencies:
+
+   ```text
+   npm install
+   ```
+
+3. Build and deploy your canisters.
+
+   ```text
+   dfx deploy
+   ```
+
+## TO DO
+
+- Set variable.
+- Check variable.
+
+## More info
+
+General background:
+
+- [Manage Canisters](https://sdk.dfinity.org/docs/developers-guide/working-with-canisters.html)
+- [Quick  Start](https://sdk.dfinity.org/developers-guide/quickstart.html)
+- [Developer's Guide](https://sdk.dfinity.org/developers-guide)
+- [Language Reference](https://sdk.dfinity.org/language-guide)
--- a/motoko/cert-var/Makefile
+++ b/motoko/cert-var/Makefile
@@ -1,0 +1,34 @@
+.PHONY: all
+all: build
+
+.PHONY: node_modules
+.SILENT: node_modules
+node_modules:
+	npm install
+
+.PHONY: build
+.SILENT: build
+build: node_modules
+	dfx canister create --all
+	dfx build
+
+.PHONY: install
+.SILENT: install
+install: build
+	dfx canister install --all
+
+.PHONY: upgrade
+.SILENT: upgrade
+upgrade: build
+	dfx canister install --all --mode=upgrade
+
+.PHONY: test
+.SILENT: test
+test: install
+	dfx canister call cert_var set 42
+
+.PHONY: clean
+.SILENT: clean
+clean:
+	rm -fr .dfx
+	rm -fr node_modules
--- a/motoko/cert-var/dfx.json
+++ b/motoko/cert-var/dfx.json
@@ -1,0 +1,34 @@
+{
+  "canisters": {
+    "cert_var": {
+      "main": "src/cert_var/main.mo",
+      "type": "motoko"
+    },
+    "cert_var_assets": {
+      "dependencies": [
+        "cert_var"
+      ],
+      "frontend": {
+        "entrypoint": "src/cert_var_assets/src/index.html"
+      },
+      "source": [
+        "src/cert_var_assets/assets",
+        "dist/cert_var_assets/"
+      ],
+      "type": "assets"
+    }
+  },
+  "defaults": {
+    "build": {
+      "packtool": ""
+    }
+  },
+  "dfx": "0.8.1",
+  "networks": {
+    "local": {
+      "bind": "127.0.0.1:8000",
+      "type": "ephemeral"
+    }
+  },
+  "version": 1
+}
--- a/motoko/cert-var/package-lock.json
+++ b/motoko/cert-var/package-lock.json
@@ -1,0 +1,5860 @@
+{
+  "name": "cert_var_assets",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@dfinity/agent": {
+      "version": "0.9.2",
+      "resolved": "https://registry.npmjs.org/@dfinity/agent/-/agent-0.9.2.tgz",
+      "integrity": "sha512-nxYagdY1zAIimgff/NbOGs4pfgNlSWw/oBR/pIHfJMIdklGGzlU69wnyqRJCL9yUCO3I/0bbxeWGzaV7aVabIw==",
+      "dev": true,
+      "requires": {
+        "base64-arraybuffer": "^0.2.0",
+        "bignumber.js": "^9.0.0",
+        "borc": "^2.1.1",
+        "buffer": "^6.0.3",
+        "buffer-pipe": "^0.0.4",
+        "js-sha256": "0.9.0",
+        "simple-cbor": "^0.4.1"
+      }
+    },
+    "@dfinity/candid": {
+      "version": "0.9.2",
+      "resolved": "https://registry.npmjs.org/@dfinity/candid/-/candid-0.9.2.tgz",
+      "integrity": "sha512-h8fzslfP5wiZLQoaAGQEEx7naaOWmfSrL8cfHRk98fiy5m3SnoQrAkIQVLr5l3tq8zFosIyYov+47lgxMTBRUA==",
+      "dev": true,
+      "requires": {
+        "buffer": "^6.0.3",
+        "buffer-pipe": "^0.0.4"
+      }
+    },
+    "@dfinity/principal": {
+      "version": "0.9.2",
+      "resolved": "https://registry.npmjs.org/@dfinity/principal/-/principal-0.9.2.tgz",
+      "integrity": "sha512-YtxM+BSYS1FlLOz/SYPyvniw1j0UbMnM99IueVLaLEgIi76gRYxV9kTewbzc57DqyCcBquvn+doZpMHHqHiuUA==",
+      "dev": true
+    },
+    "@discoveryjs/json-ext": {
+      "version": "0.5.5",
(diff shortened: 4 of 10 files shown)


matthewhammer:
As I mentioned in the comments of #118, I am creating this example as a derivative of `random_maze`.  My rationale is that @rossberg and @crusso both favor minimal examples, and as far as I can see, `random_maze` is currently the most minimal example with a front-end asset canister, and javascript code.

cc @chenyan-dfinity @crusso

matthewhammer:
Update: Just chatted with @chenyan-dfinity and I have some pointers to guide the next steps.

- [Example code to follow on client side.](https://github.com/dfinity/agent-js/blob/d861138eab0de6f8425daea420d3e46ec07fa01d/e2e/node/basic/basic.test.ts#L16)
- [Relevant text from spec, about the certified data format.](https://sdk.dfinity.org/docs/interface-spec/index.html#system-api-certified-data)

Thanks Yan!

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+      ...(new Uint8Array(rawTime) || []),
+    ]),
+  )[0];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const time = Number(decoded as any) / 1e9;
+  // The diff between decoded time and local time is within 5s
+  expect(Math.abs(time - now) < 5).toBe(true);
+
+  const pathData = [te.encode('canister'),
+                    te.encode("to do -- canister id"),
```
matthewhammer, line 39: @chenyan-dfinity What's the canonical way of getting the canister Id, textually (?), of the `cert_var` canister?

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+  const decodedData = IDL.decode(
+    [IDL.Nat32],
+    new Uint8Array([
+      ...new TextEncoder().encode('DIDL\x00\x01\x7d'),
+      ...(new Uint8Array(rawData) || []),s
+    ]),
+  )[0];
+  // the certified data should be the same as the value
+  // (to do -- modulo the byte ordering of how we represent it...?)
+  expect(decodedData == resp.value).toBe(true);
```
matthewhammer, line 51: @chenyan-dfinity I may need to do something byte-level here, with a loop, if the byte order does not match.  Ideally, I'd avoid that by having the byte order that my canister chooses match whatever makes sense here, for the IDL encoding of `Nat32`.  Does that seem right?

motoko/cert-var/src/cert_var_assets/src/index.js
chenyan-dfinity: https://github.com/dfinity/agent-js/blob/main/packages/principal/src/index.ts#L76

motoko/cert-var/src/cert_var_assets/src/index.js
chenyan-dfinity: Yeah, nat32 uses LE. You can use `IDL.Nat32.encodeValue(nat32_number)` to get the blob.

chenyan-dfinity:
Now that I know a bit more, I think the proof-of-concept doesn't have to be a frontend app. It can be a Rust agent trying to certify the return from a query method as well. If the frontend is too complicated, you can start with ic-repl or even your ic-terminal tool.

matthewhammer:
> I think the proof-of-concept doesn't have to be a frontend app. It can be a Rust agent trying to certify the return from a query method as well.

Yes, but perhaps the more ways we show, the better?  If someone wants to copy and use the JS version of the checks, it would be nice to demonstrate them here, right? (along with whatever other client versions we want to show)

motoko/cert-var/README.md
```diff
@@ -0,0 +1,54 @@
+# Certified Variable
+
+![Compatibility](https://img.shields.io/badge/compatibility-0.7.0-blue)
+[![Build Status](https://github.com/dfinity/examples/workflows/motoko-cert-var-example/badge.svg)](https://github.com/dfinity/examples/actions?query=workflow%3Amotoko-cert-var-example)
+
+The example shows a cryptographically certified variable.
```
crusso, line 6:
```suggestion
The example demonstrates the use of a single cryptographically certified variable, as supported by the Internet Computer. 
```
And maybe ad links to the IC Public Spec section(s) about them (either here or below) and perhaps the video explaining their use.

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -0,0 +1,53 @@
+import { Actor, HttpAgent } from '@dfinity/agent';
+import { Certificate } from '@dfinity/agent';
+import { cert_var } from '../../declarations'
+
+document.getElementById("certifyBtn").addEventListener("click", async () => {
+  const newVal = BigInt(document.getElementById("newValue").value);
+  const _ = await cert_var.set(newVal);
+  const resp = await cert_var.get();
+
+  const agent = await Promise.resolve(new HttpAgent({ host: 'http://127.0.0.1:' + port, identity })).then(
```
matthewhammer, line 10: @chenyan-dfinity How do I determine `port` here?  Should I hardcode a port to go with the hardcoded URL, like 8080?

matthewhammer: Oh, but I also need to know `identity` as well.

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -0,0 +1,67 @@
...
+  const agent = await Promise.resolve(new HttpAgent({ host: 'http://127.0.0.1:' + port })).then(
+      async agent => {
+          await agent.fetchRootKey();
+          agent.addTransform(makeNonceTransform());
+          return agent;
+      },
+  );
+  const now = Date.now() / 1000;
+
+  const cert = new Certificate(resp, agent);
```
matthewhammer, line 21: @chenyan-dfinity 

When I try this out locally, I get this in the browser console:
```
decoder.js:570 Uncaught (in promise) Error: No valid result
    at Decoder._decode (decoder.js:570)
    at Decoder.decodeFirst (decoder.js:577)
    at Module.decode (cbor.js:94)
    at new Certificate (certificate.js:75)
    at HTMLButtonElement.<anonymous> (index.js:21)
 ```

Comment (chenyan-dfinity) [reply motoko/cert-var/src/cert_var_assets/src/index.js:21]: Certificate expects a blob, so it should be `resp[1][0]`, meaning the second return value and unwrapping the opt.

Comment (matthewhammer) [reply motoko/cert-var/src/cert_var_assets/src/index.js:21]: No, nevermind.  When I use the dev tools, I see that it is not null.

File: motoko/cert-var/src/cert_var_assets/src/index.js [hunk 6]
```diff
@@ -0,0 +1,66 @@
...
+  // to do -- From where to get the port and host, in general?
+  const port = 8000;
+  const agent = await Promise.resolve(new HttpAgent({ host: 'http://127.0.0.1:' + port })).then(
+      async agent => {
+          await agent.fetchRootKey();
+          agent.addTransform(makeNonceTransform());
+          return agent;
+      },
+  );
+  const cert = new Certificate(resp.certificate[0], agent);
```
matthewhammer, line 19: @chenyan-dfinity Can you take a look at this line?

Unlike before, I think I'm now supplying the blob that this constructor expects to get.

However, I get this error now:
```
cbor.js:86 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'byteLength')
    at Module.decode (cbor.js:86)
    at new Certificate (certificate.js:75)
    at HTMLButtonElement.<anonymous> (index.js:19)
```

Using a breakpoint, I can see that the blob exists and is 442 bytes long, for what it's worth.

BTW, when I leave off the agent, I get an error about the default agent not being defined in this context.  I tried to vary that when I wasn't sure about that being cause, but it didn't help.

chenyan-dfinity:
You need to run `npm install` first. I bumped the agent version. It should work after that.

matthewhammer:
> Another random thought: Since the certification logic is application-specific, we can use the custom section to specify how we can verify the query result. The client (frontend or dfx) can use the custom section to reproduce the verification. Not sure what kind of DSL we need for this, but then the client will just be an interpreter for this DSL.

This idea appeals to me, assuming we can agree about the custom section format, and the DSL, etc.

@nomeata had [an internal document](https://docs.google.com/document/d/1A2SztwkIemVwH6zzBBecqLee69Xo2w9vft27x6zGZF0) that proposed a similar idea, along side other related ideas.

matthewhammer:
Some manual testing exposed bugs in `main.mo` that I fixed above.  Now things seem solid, even for large values.

Looking at the other examples here, I don't see that they do any testing of the JS code in the frontend, and just test the backend in isolation.

What's the preferred way to do that here, I wonder?

matthewhammer:
Okay, there seems to be some consensus about not doing CI-based e2e testing here, for now.

motoko/cert-var/README.md
```diff
@@ -0,0 +1,54 @@
...
+   npm install
+   ```
+
+3. Build and deploy your canisters.
+
+   ```text
+   dfx deploy
+   ```
+
+## TO DO
```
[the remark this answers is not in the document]
matthewhammer: Oops!  Nope -- thanks!

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -0,0 +1,77 @@
...
+  const readState = { certificate: new Uint8Array(resp.certificate[0]) };
+  const cert = new Certificate(readState, agent);
+
+  // Check: Certificate verifies.
+  if(!(await cert.verify())) {
+    log.innerText = "Failure: Certification verification failed.";
+    return;
+  }
+  
+  const te = new TextEncoder();
```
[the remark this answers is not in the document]
matthewhammer: Good point!

With your encouragement, I took an initial pass at authoring that doc, [here](https://github.com/dfinity/examples/pull/119/commits/eaf0b2f232f8d736e2b2164a5ee89448d6fab08e).

WDYT?

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -25,15 +25,67 @@ document.getElementById("setBtn").addEventListener("click", async () => {
...
+
+    https://sdk.dfinity.org/docs/interface-spec/index.html#system-api-certified-data
+
+    Before we trust the response from our apparently "true" canister,
+    we interrogate it, and verify its authenticity:
+
+    We do four checks below:
+
+    1. verify system certificate.
+    2. check system certificate timestamp is not "too old".
```
chenyan-dfinity, line 56: It's worth pointing out that query call doesn't necessarily return the latest state. Getting an old timestamp doesn't mean you are under attack. It may mean that the node responding to the query call hasn't caught up with the latest state yet.

motoko/cert-var/src/cert_var_assets/src/index.js
```diff
@@ -25,15 +25,67 @@ document.getElementById("setBtn").addEventListener("click", async () => {
...
+    3. check canister ID in system certificate.
+    4. check response matches witness.
+
+    For steps 2, 3 and 4, we access data from the certificate (Blob).
+    The Certificate class from the agent-js library provides a way to
+    access those items using their paths, like a filesystem, each addressing
+    a Blob, encoding something.
+
+    In the case of time and our data, the encodings are each Candid.
```
chenyan-dfinity, line 65: The spec says `time` is using the LEB128 encoding, and our certified data is represented in little endian. Ideally, we should use a proper library to decode these numbers. To prevent an extra dependency, we take advantage of the fact that the Candid value encoding of `nat` and `nat32` happens to use the same representation we need.

motoko/cert-var/src/cert_var_assets/src/index.js
nomeata: Well, if it is older than a few minutes, something is likely odd, and it seems reasonable to treat that as failure? After all, it could be a node that maliciously and intentionally sends you old data.

PR #127, opened 2021-12-11
ROSETTA1-166 A Rust canister that transfers funds using ICP Ledger API 

MarioDfinity:
Overview
This PR adds an simple example of Ledger usage. The Tokens Transfer canister allows to check the canister balance and transfer tokens from its account to other accounts.


rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Hash)]
+pub struct TransferArgs {
+    amount: Tokens,
+    to_account: Principal,
+    to_subaccount: Option<Subaccount>,
+}
+
+// returns the memo as transaction id
```
roman-kashitsyn, line 56: I think we should return the block index as well. I'm not sure why would we return memo...

rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+#[init]
+#[candid_method(init)]
+fn init(conf: Conf) {
+    CONF.with(|c| c.replace(conf));
+}
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Hash)]
+pub struct TransferArgs {
+    amount: Tokens,
+    to_account: Principal,
```
roman-kashitsyn, line 52:
```suggestion
    to_principal: Principal,
```
I find it confusing: we identify accounts with 32-byte arrays, not principals.

rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+fn hash_transfer_input(args: &TransferArgs, conf: &Conf) -> u64 {
+    let mut hasher = DefaultHasher::default();
+    std::time::SystemTime::now().hash(&mut hasher);
+    id().hash(&mut hasher);
+    args.hash(&mut hasher);
+    conf.hash(&mut hasher);
+    hasher.finish()
+}
```
roman-kashitsyn, line 87: I'm not sure I understand the purpose of constructing the memo this way.
We could use a fixed memo or at least provide an explanation why we do it this way, otherwise people might assume that this is the way you should always construct memos.

rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use crate::errors::TransferError;
+
+#[cfg(test)] mod candid_util;
+mod errors;
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Hash, PartialEq)]
+pub struct Conf {
+    ledger_canister_id: Principal,
+    subaccount: Option<Subaccount>,
```
roman-kashitsyn, line 25: Whose subaccount it is? Subaccount that this canister should always use?
I find it a bit weird: why would you fix a single subaccount? If you only allow one, just use the default one.

rust/tokens_transfer/src/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn from(e: (RejectionCode, String)) -> Self {
+        TransferError::CallError(format!("rejection_code:{:?} error:{}", e.0, e.1))
+    }
+}
+
+impl From<ic_ledger_types::TransferError> for TransferError {
+    fn from(e: ic_ledger_types::TransferError) -> Self {
+        TransferError::LedgerError(format!("{}", e))
+    }
+}
```
roman-kashitsyn, line 29: Not sure it makes sense to have an elaborate error data type in the example, given that this canister is not designed as an API for other canisters. Returning an error as a string is fine IMO.

rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[test]
+fn check_candid_interface_compatibility() {
+    candid::export_service!();
+    let actual_interface = __export_service();
+    let mut actual = tempfile::NamedTempFile::new().expect("Failed to create a temporary file");
+    write!(actual, "{}", actual_interface).unwrap();
+    let expected = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("src").join("tokens_transfer.did");
+    candid_util::check_candid_interface_compatibility(&expected, &actual.path());
+}
```
roman-kashitsyn, line 103: Not sure we need to track backward compatibility here: this check is _really_ important for heavily-used production canisters, but for examples it's just unnecessary complexity.

rust/tokens_transfer/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. start the local replicate with `dfx start`
+2. deploy the Ledger canister locally. Add some tokens to your account (`dfx identity get-principal`) in the initialization parameters of the Ledger canister.
+```bash
+# MINTING_ACCOUNT_ID_HEX and ACCOUNT_ID_HEX are the hex representation
+# of the minting principal and your principal respectively
+read -r -d '' ARGS <<EOM
+(record {
+     minting_account="${MINTING_PRINCIPAL_HEX}";
+     initial_values=vec { record { "${YOUR_PRINCIPAL_HEX}"; record { e8s=10_000_000_000 } }; };
```
roman-kashitsyn, line 30: Are you sure those are _principals_ and  not _accounts ids_?

rust/tokens_transfer/README.md
```diff
@@ -0,0 +1,71 @@
...
+}, )
+EOM
+dfx deploy --argument "${ARGS}" tokens_transfer
+```
+4. transfer some funds to the Tokens Transfer canister
+```bash
+# TOKENS_TRANSFER_PRINCIPAL_ID_BYTES is the vec nat8 representation of the tokens transfer canister
+read -r -d '' ARGS <<EOM
+(record {
+  to=${TOKENS_TRANSFER_PRINCIPAL_ID_BYTES};
```
roman-kashitsyn, line 53: I'm pretty sure this is an account id, not principal.

rust/tokens_transfer/src/lib.rs
MarioDfinity: I think it makes sense in the context of an example to show that there are several subaccounts for each account.

rust/tokens_transfer/src/lib.rs
MarioDfinity: The idea is to also show how to create account id from principals.

rust/tokens_transfer/src/lib.rs
MarioDfinity: Removed the memo and returned BlockIndex

rust/tokens_transfer/src/lib.rs
MarioDfinity: Removed.

rust/tokens_transfer/src/errors.rs
MarioDfinity: Removed and returning a string.

rust/tokens_transfer/README.md
MarioDfinity: Changed to account ids.

rust/tokens_transfer/README.md
MarioDfinity: Changed to account ids.

roman-kashitsyn:
As `ic-ledger-types` has to depend `ic-cdk` anyways (doesn't compile without it for unclear reasons), I decided to add `account_balance` and `transfer` async functions to make the client code a bit nicer: https://github.com/dfinity/cdk-rs/pull/168/commits/83bf346160f4f89cfded9545585dd8eacf0c0b1e
Feel free to use them in your example.

MarioDfinity:
I think it's a good idea to expose those functions for client code. I've changed the example to use ic_ledger_types::transfer and removed my utility function. Thank you.

MarioDfinity:
@roman-kashitsyn I renamed to_account to to_principal to better describe what the field is and to avoid confusion with AccountIdentifiers.

rust/tokens_transfer/src/lib.rs
MarioDfinity: I applied the change with the last commit

rust/tokens_transfer/src/lib.rs
MarioDfinity: I pushed a comment and add a small description and a link to the Ledger doc.

rust/tokens_transfer/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+The canister expects three arguments:
+1. `ledger_canister_id`: the canister id of the ledger canister
+2. `subaccount`: the optional subaccount of the canister account from which tokens will be withdrawn
+3. `transaction_fee`: a constant representing the transaction fee of the ledger
+
+
+## Test Locally
+
+1. start the local replicate with `dfx start`
```
roman-kashitsyn, line 22: We should probably link to this doc https://github.com/dfinity/ic/tree/master/rs/rosetta-api/ledger_canister#deploying-locally to explain how to build and install the ledger.

rust/tokens_transfer/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+            memo: Memo(0),
+            amount: args.amount,
+            fee: conf.transaction_fee,
+            from_subaccount: conf.subaccount,
+            to: AccountIdentifier::new(&args.to_principal, &to_subaccount),
+            created_at_time: None,
+        }
+    });
+    ic_ledger_types::transfer(ledger_canister_id, transfer_args).await
+        .map_err(|e| format!("ledger transfer error {:?}", e))?
```
roman-kashitsyn, line 65:
```suggestion
        .map_err(|e| format!("failed to call ledger: {:?}", e))?
```

rust/tokens_transfer/src/tokens_transfer.did
```diff
@@ -0,0 +1,32 @@
...
+type TransferError = variant {
+    AccountError: text;
+    CallError: text;
+    LedgerError: text;
+};
+
+type TransferResult = variant {
+    Ok: Memo;
+    Err: TransferError;
+};
```
roman-kashitsyn, line 28:
```suggestion
type TransferResult = variant {
    Ok: Memo;
    Err: text;
};
```

rust/tokens_transfer/README.md
MarioDfinity: I've replaced the first two points with a link to that doc.

roman-kashitsyn (approved):

PR #128, opened 2021-12-15
ROSETTA-1-176 add Ledger get_blocks_pb example

MarioDfinity:
This is an example of how to get the Ledger blocks via the get_blocks_pb. It can be used as template by other devs to understand how they can fetch the blocks using protocol buffers instead of candid.

### DIFF
--- a/rust/get-ledger-blocks/src/main.rs
+++ b/rust/get-ledger-blocks/src/main.rs
@@ -1,0 +1,70 @@
+use std::error::Error;
+use ic_agent::Agent;
+use ic_agent::agent::http_transport::ReqwestHttpReplicaV2Transport;
+use ic_agent::ic_types::Principal;
+use prost::Message;
+use structopt::StructOpt;
+use url::Url;
+use ledger::GetBlockRequest;
+use crate::ledger::get_blocks_response::GetBlocksContent;
+use crate::ledger::{Block, GetBlocksResponse};
+
+pub mod ledger {
+    include!("gen/ledger.rs");
+}
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "get-ledger-blocks", about = "An example of how to get blocks from the ledger")]
+struct Opt {
+    #[structopt(short, long)]
+    url: Url,
+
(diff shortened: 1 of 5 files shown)


MarioDfinity:
I copied the relevant parts of the  ledger proto files inside ledger.proto.

rust/get-ledger-blocks/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+        .call()
+        .await?;
+
+    // decode the response as protobuf message
+    let res = GetBlocksResponse::decode(res.as_slice()).unwrap().get_blocks_content.unwrap();
+
+    // check if the response is an error or the encoded blocks
+    let encoded_blocks = match res {
+        GetBlocksContent::Blocks(blocks) => blocks.blocks,
+        GetBlocksContent::Error(e) => panic!("Error fetching blocks {}", e),
```
eust-dfinity, line 60: I'm afraid that this example is too simplistic to be actually useful. Most blocks are in the archive, so you really need to fetch the archive index first and based on that request blocks

MarioDfinity: Should I wait for the fetch blocks API implementation and only then push this example? I don't think I can query a range include archives without it.

eust-dfinity: Yeah, using the new endpoint would be best. But we do query the index manually in rosetta right now:
https://gitlab.com/dfinity-lab/public/ic/-/blob/master/rs/rosetta-api/src/ledger_client.rs#L1096

Also, there is an endpoint for querying a single block which behaves just like the new interface will:
https://gitlab.com/dfinity-lab/public/ic/-/blob/master/rs/rosetta-api/src/ledger_client.rs#L950-973

Although we were planning to deprecate both and switch to the one @roman-kashitsyn has been working on...
@roman-kashitsyn what do you think?

MarioDfinity:
There is no need to add this example given the discussions around protobuf and the new fetch_blocks API.

PR #134, opened 2022-01-17
feat: template for motoko/rust examples: project hello

crusso:
This PR tries to create a single, coherent example of the same dapp in both Rust in Motoko.

The example is just the default project created with `dfx new (--type rust)` and has both a Motoko (Rust) backend and frontend.
(Unfortunately, none of the existing dual language examples are either equivalent or simple enough.)

Each project has:

- [x] A tutorial style README for setting up, building and running the sample
- [x] A link to the same example in the other language (if possible).
- [x] Makefile (to be used for testing by CI)
- [x] GH actions so it can be tested against newer versions of dfx (leveraging the Makefile)
- [x] Docs onepager, c.f. https://github.com/dfinity/docs/pull/659
- [x] RECIPE.md outlining steps required for adding a new dual language example.

Issues:
- How does one test the asset canister/frontend? The only existing example is out of date and commented out.
- Might be nicer if we could have the adoc in this examples repo, not the docs repo.


bogwar (approved):
lgtm

ADDING_AN_EXAMPLE.md
```diff
@@ -0,0 +1,59 @@
+# How to Add a New Example
+
+Ideally, each example should come with both Rust and Motoko implementations,
+implementing the same Candid interface (and, ideally, semantics).
```
chenyan-dfinity, line 4: One exception is the variant type. Rust prefers to capitalize the first character of variant tags, while Motoko prefers to use lower case.

PR #139, opened 2022-01-25
Kyle/motoko unit tests

krpeacock:
**Overview**
This demonstrates a pattern for running unit tests on Motoko modules

**Requirements**
We should have a pattern for running a simple test suite with pass / fail conditions

**Recommended Solution**
This is a pattern that used to be in the Motoko base library as an example


dfx-json:
thank you for your contribution. as this PR is quite outdated, I will be closing it. if you would still like to submit your changes, please do so in a new PR. thank you for your understanding.

PR #140, opened 2022-01-26
DAO example in Motoko

chenyan-dfinity:


crusso:
Will this be merged with #142. It'll need a README (maybe just symlink the other one if that works)

Do the rust/motoko examples implement exactly the same interface?

Can they use your ic-repl test to the rust example?

motoko/dao/src/Main.mo
```diff
@@ -0,0 +1,221 @@
...
+            switch (await execute_proposal(proposal)) {
+            case (#ok) { update_proposal_state(proposal, #succeeded); };
+            case (#err(err)) { update_proposal_state(proposal, #failed(err)); };
+            };
+        };
+    };
+
+    /// Execute the given proposal
+    func execute_proposal(proposal: Types.Proposal) : async Types.Result<(), Text> {
+        // unimplemented until raw call is supported
```
crusso, line 200: Ha!

motoko/dao/tests/proposal.test.sh
```diff
@@ -0,0 +1,148 @@
+#!ic-repl
+load "prelude.sh";
```
crusso, line 2: ic-repl looks nice - I should actually use it...

crusso: But it also looks very sequential - are all calls made synchronous?

motoko/dao/src/Types.mo
```diff
@@ -0,0 +1,65 @@
+import Result "mo:base/Result";
```
crusso, line 1: Does this mean Rust and Motoko will have different candid interface with #Err vs #err?

motoko/dao/tests/proposal.test.sh
chenyan-dfinity: All calls have an await, so the script is sequential. Do you have a use case to make concurrent calls?

motoko/dao/src/Types.mo
chenyan-dfinity: Yes

motoko/dao/src/Types.mo
```diff
@@ -0,0 +1,65 @@
...
+import List "mo:base/List";
+import Principal "mo:base/Principal";
+
+module {
+  public type Result<T, E> = Result.Result<T, E>;
+  public type Account = { owner : Principal; tokens : Tokens };
+  public type Proposal = {
+    id : Nat;
+    votes_no : Tokens;
+    voters : List.List<Principal>;
```
chenyan-dfinity, line 13: Besides `Err` vs `err`, this is another difference in the did file. Rust is using `vec principal`, and Motoko uses `opt record { principal; list }`.

The Rust implementation can use the same ic-repl scripts, except for the above differences.

ggreif (approved):
Go for it!

PR #164, opened 2022-02-23
Encrypted notes ci

nikolay-komarevskiy:


aterga:
The Motoko canisters are failing in the CI pipelines due to new DFX behavior. In particular, the most recent stable version of DFX seems to be 0.9.2, introducing automatic [stable safety checking](https://github.com/dfinity/sdk/blob/458abd206c30ecd98e5448e35803ddfbd1344063/CHANGELOG.adoc#092). 

At the end of the deployment process, we get the following issue: 
```
Installing canisters...
WARNING!
Stable interface compatibility check failed for canister 'encrypted_notes_motoko'.
Upgrade will either FAIL or LOSE some stable variable data.

/private/tmp/examples/motoko/encrypted-notes-dapp/.dfx/local/canisters/encrypted_notes_motoko/encrypted_notes_motoko.old.most:3.41-3.50: type error [M0029], unbound type PublicKey
```

This raises two questions: 

1. The warning is correct: our canister uses unstable fields, but they are serialized via stable arrays in an upgrade scenario. How can one convince DFX that this is okay? 

    The interactive prompt is a bit weird. DFX produces the following .most file: 
    ```
    type EncryptedNote = {encrypted_text : Text; id : Nat};
    type PrincipalName = Text;
    type StableUserStoreEntry = (Principal, PublicKey, DeviceAlias, ?Ciphertext);
    actor {
      stable var nextNoteId : Nat;
      stable var stable_notesByUser : [(PrincipalName, [EncryptedNote])];
      stable var stable_users : [StableUserStoreEntry]
    };
    ```

2. For some reason, DFX successfully resolves the type `StableUserStoreEntry` from `user_store.mo` but fails to resolve `PublicKey` from `types.mo`.

chenyan-dfinity:
@crusso This seems like a bug in `--stable-type`. The type definition of `PublicKey` is somehow not in the `.most` file.

crusso:
@chenyan-dfinity thanks, I'll take a look.

crusso:
Yeah, this is definitely a bug in Motoko:
https://github.com/dfinity/motoko/issues/3128

Until that is fixed, you can try merging the workaround here #170 (inlining the type definition).

The problem is that Motoko isn't generating all the types it needs to make the .most file self-contained because it fails to include the type definitions that are mentioned only by the bodies of referenced type definitions (here PublicKey).

aterga:
Thanks a lot for a quick response, @crusso !
Cc @nikolay-komarevskiy

motoko/encrypted-notes-dapp/README.md
```diff
@@ -0,0 +1,354 @@
...
+- [Troubleshooting](#troubleshooting)
+  - [Building/deployment problems](#buildingdeployment-problems)
+  - [Login problems](#login-problems)
+- [dfx.json file structure](#dfxjson-file-structure)
+- [Local memory model](#local-memory-model)
+- [Further Reading](#further-reading)
+- [Acknowledgments](#acknowledgments)
+## Encrypted notes
+
+Encrypted Notes is an experimental dapp for authoring and storing confidential information on the Internet Computer (IC) in the form of short pieces of text. Users can create and access their notes via any number of automatically synchronized devices authenticated via [Internet Identity (II)](https://smartcontracts.org/docs/ic-identity-guide/what-is-ic-identity.html). Notes are stored confidentially thanks to the end-to-end encryption performed by the dapp’s frontend.
```
yvonneanne, line 25:
```suggestion
Encrypted Notes is an example dapp for authoring and storing confidential information on the Internet Computer (IC) in the form of short pieces of text. Users can create and access their notes via any number of automatically synchronized devices authenticated via [Internet Identity (II)](https://smartcontracts.org/docs/ic-identity-guide/what-is-ic-identity.html). Notes are stored confidentially thanks to the end-to-end encryption performed by the dapp’s frontend.
```

motoko/encrypted-notes-dapp/README.md
```diff
@@ -0,0 +1,354 @@
...
+---
+&nbsp;
+
+## Disclaimer: please read carefully
+
+This is an _example dapp_ that demonstrates the potential of building _canisters_ for the IC. Please do not use this code in production and/or scenarios in which sensitive data could be involved. While this dapp illustrates end-to-end encryption, there are **several open security issues** that should be addressed before the dapp could be considered production-ready:
+
+- The frontend re-uses the generated public- and private-key pair for every identity in the same browser. This key pair should be unique per principal.
+- The public/private key pair should not be managed by the web browser at all. [WebAuthn](https://en.wikipedia.org/wiki/WebAuthn) should be used to push the key management to the operating system.
+- Currently, the notes are encrypted deterministically. However, this is not a conceptual requirement of the dapp. Symmetric encryption should be replaced by a proper AES-GCM, using per-note unique initialization vectors.
```
yvonneanne, line 38:
```suggestion
```
I changed that before my holidays :)

motoko/encrypted-notes-dapp/README.md
```diff
@@ -0,0 +1,354 @@
...
+- Integer overflows are possible in the Rust canister, e.g., for `NEXT_NOTE`.
```
yvonneanne, line 39:
```suggestion
- Integer overflows are possible in the Rust canister, e.g., for `NEXT_NOTE`. 
- Users may lose their notes if they accidentally clean the browser data (localStorage) and no other device is synced to the dapp.
- Lack of key update: Given that the key used to encrypted the notes is never refreshed, the privacy of the data is no longer guaranteed if an attacker learns this key (for instance, by corrupting the local storage in one of the connected devices).
```
Suggested by Michel: https://docs.google.com/document/d/1czp7C-oLBsuk2mwASHIPOeYmpzVVtdRXAhOBo9e4kd8/edit

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+    assert!(is_user_registered(user));
+    assert!(is_id_sane(note_id));
+
+    let user_str = user.to_string();
+    // shared ownership borrowing
+    NOTES_BY_USER.with(|notes_ref| {
+        let mut writer = notes_ref.borrow_mut();
+        if let Some(v) = writer.get_mut(&user_str) {
+            v.retain(|item| item.id != note_id);
+        }
```
yvonneanne, line 201: don't we want to reduce the counter in this case?

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+    id: u128,
+    encrypted_text: String,
+}
+
+/// There can only be one Type in stable storage at a time.
+/// We use this struct to represent the full CanisterState
+/// So we can serialize it to stable storage.
+#[derive(Clone, CandidType, Serialize, Deserialize)]
+struct CanisterState {
+    counter: u128,
```
yvonneanne, line 50:
```suggestion
    // number of notes across all users (to avoid running out of memory)
    counter: u128,
```

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+    }
+}
+
+// WASM is single-threaded by nature. [RefCell] and [thread_local!] is used vice unsafe.
+// This is to ensure that the canister state can be used throughout.
+// Your other option here is to avoid [thread_local!] and use a [RefCell<RwLock/Mutex/Atomic>].
+// Here we use [thread_local!] because it is simpler.
+thread_local! {
+
+    // Define dapp limits - important for security assurance
```
yvonneanne, line 75: My back of the envelope calculation gives 1GB of notes, is this what you had in mind?

motoko/encrypted-notes-dapp/src/frontend/src/lib/crypto.ts
```diff
@@ -0,0 +1,294 @@
...
+      this.deviceAlias = deviceAlias;
+    } else {
+      this.deviceAlias = uuidv4();
+      window.localStorage.setItem('deviceAlias', this.deviceAlias);
+    }
+    console.log('deviceAlias: ' + this.deviceAlias);
+  }
+
+  private publicKeyBase64: string | null = null;
+  private privateKey: CryptoKey | null = null;
```
yvonneanne, line 22:
```suggestion
  // Used to decrypt the symmetric secretKey, which will be stored by the dapp encrypted with the the publicKey
  private privateKey: CryptoKey | null = null;
```

motoko/encrypted-notes-dapp/src/frontend/src/lib/crypto.ts
```diff
@@ -0,0 +1,294 @@
...
+  private publicKey: CryptoKey | null = null;
+  private secretKey: CryptoKey | null = null;
+  private secret: string | null = null;
```
yvonneanne, line 25:
```suggestion
  // Symmetric AES key, used to encrypt and decrypt the notes stored in the dapp
  private secretKey: CryptoKey | null = null;
  private secret: string | null = null;
    // Used to decrypt the symmetric secretKey, which is stored by the dapp encrypted with the the publicKey
  private privateKey: CryptoKey | null = null;
    // Used to encrypt the symmetric secretKey for each device associated with the current principal, to be stored by the dapp in encrypted form
  private publicKey: CryptoKey | null = null;
  private publicKeyBase64: string | null = null;
```

motoko/encrypted-notes-dapp/README.md
```diff
@@ -0,0 +1,354 @@
...
+This project serves as a simple (but not too simple) example of a dapp, which uses Motoko and Rust as backend and Svelte as frontend.
+
+---
+&nbsp;
+
+## Disclaimer: please read carefully
+
+This is an _example dapp_ that demonstrates the potential of building _canisters_ for the IC. Please do not use this code in production and/or scenarios in which sensitive data could be involved. While this dapp illustrates end-to-end encryption, there are **several open security issues** that should be addressed before the dapp could be considered production-ready:
+
+- The frontend re-uses the generated public- and private-key pair for every identity in the same browser. This key pair should be unique per principal.
```
yvonneanne, line 36:
```suggestion
- The frontend re-uses the generated public- and private-key pair for every identity in the same browser.  In a better implementation this key pair should be unique per principal.
```
https://docs.google.com/document/d/1czp7C-oLBsuk2mwASHIPOeYmpzVVtdRXAhOBo9e4kd8/edit

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
yvonneanne: Actually, we don't seem to use the counter, or am I missing something?

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
aterga: Not in the current design, because the counter is used as a source of unique note ids.

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
aterga: During canister upgrades, the `counter` field of `CanisterState` is simply a stable representation of the value stored in `NEXT_NOTE`.

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
aterga: My calculations are:
1. Each user may store ~1 GB worth of notes
2. Each user may store up to ~2 MB for device management (negligible)

Since we permit up to 1000 users, our constraints do not ensure that canister limit of 4 GB is respected.

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
yvonneanne: Could you add a comment about that, please? (also in Motoko)

motoko/encrypted-notes-dapp/src/encrypted_notes_rust/src/lib.rs
yvonneanne: Shouldn't we limit the number of notes to a smaller value to ensure the canister does not run out of memory?
E.g. 5000 notes per user still seems plenty :)

aterga: Addressed in https://github.com/dfinity/examples/pull/164/commits/dceba856d20ed24916b34a33819968896946557b

aterga (approved):

nikolay-komarevskiy:
Publish Encrypted Notes dapp

PR #167, opened 2022-02-25
Defi example dapp

tim-dfn:
Edit: (Eero) Links to live deployment have been added.


krpeacock:
Running through the installation from the Readme from a fresh clone of the repo, the frontend is not loading successfully. Specifically, `bundle.js` is returning

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width,initial-scale=1" />

    <title>Defi Dapp</title>

    <link rel="icon" type="image/png" href="/favicon.png" />
    <link
      href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400&display=swap"
      rel="stylesheet"
    />

    <link rel="stylesheet" href="/global.css" />
    <link rel="stylesheet" href="/build/bundle.css" />

    <script defer src="/build/bundle.js"></script>
  </head>

  <body></body>
</html>```

**Comment by krpeacock** [comment]
Additionally, the pattern of separating `frontend_assets` into a different directory, and outputting the bundle from `frontend` into it, is odd. It's not wrong per se, but it's a different pattern than we typically follow elsewhere. 

I would ordinarily keep the assets in an `assets` directory under frontend, and then output any bundled code into a `dist` directory (or whatever convention is common for Svelte). The `dist` directory would be git ignored, and then I would include both directories with `"source": ["dist",  "src/frontend/assets"]`

**Comment by tim-dfn** [comment]
Thanks for the feedback. The instructions are in a weird state atm since they are intended to be correct after merging. Do you still face the issue if you follow these instructions:
```bash

git clone https://github.com/Bownairo/examples.git
git checkout defi-example
git submodule update --init --recursive
make install
```
I agree we should move the assets.

Bownairo:
> Additionally, the pattern of separating frontend_assets into a different directory, and outputting the bundle from frontend into it, is odd.

We wanted to symlink the `frontend` folder to share it for Rust and Motoko, but `dfx` protects against loading assets from outside of the root directory. We fought a bit with how everything holds together with the links, but I think we should be able to keep this linking and fit into the `frontend/assets` structure you lay out.

jason-icp (approved):
I will approve with the following suggestions:

1. add troubleshooting section to README with suggestion from tim, which was to run the following commands:
```
git clone https://github.com/Bownairo/examples.git
cd examples
git checkout defi-example
git submodule update --init --recursive
cd motoko/defi
make install
```
2. Update the folder structure to one of the more commonly used patterns at a later date.

crusso:
Sorry folks, I never noticed the review request. It would be good if the example had CI and some formatting changes to the motoko code (to adhere to guidelines), but I'd just merge it now and open an issue to address the other stuff later (since there seems to be some pressure to get this out).

PR #175, opened 2022-03-02
Change comment about canister memory limitation to 8GB instead of 4 GB

yvonneanne:
See https://forum.dfinity.org/t/increased-canister-smart-contract-memory/6148/61 and https://www.dfinitycommunity.com/internet-computer-tech-a-breakdown/


motoko/encrypted-notes-dapp/src/encrypted_notes_motoko/main.mo
```diff
@@ -22,7 +22,7 @@ import UserStore "user_store";
 // Bind the caller and the initializer
 shared({ caller = initializer }) actor class() {
 
-    // Currently, a single canister smart contract is limited to 4 GB of storage due to WebAssembly limitations.
+    // Currently, a single canister smart contract is limited to * GB of storage due to WebAssembly limitations.
```
sesi200, line 25:
```suggestion
    // Currently, a single canister smart contract is limited to 8 GB of storage due to WebAssembly limitations.
```
typo?

yvonneanne: @ielashi is there a good way to rephrase this so we don't have to update it again in a few months? 

E.g., 'At the time of writing this canister, the storage was limited to 8GB. In the meantime, larger amounts have been enabled, yet it is a good idea to have an upper bound in place to restrict cycle consumption.'

ielashi: This is actually 4GiB given that we only store the data on the heap in this dapp, and I don't think we'll ever realistically increase the size of the heap - at least not anytime soon.

yvonneanne: I seem to have really mixed up the different memory parts... sorry for the confusion, Islam and Severin!

PR #176, opened 2022-03-09
Rename the rust dao example to match the motoko name

akhi3030:
This is needed to be compliant with the requirements that the sample dapps in the different languages have the same name.


chenyan-dfinity (approved):

chenyan-dfinity (changes requested):
Wait, we need to update the dfx.json as well.

rstout:
There are many uses of `basic_dao` in the code (e.g. Makefile), do we want to update those as well?

chenyan-dfinity:
Yep, those as well. Otherwise, it won't build.

rstout:
Also, why change this from `basic_dao` to `dao`? We should be explicit that this is a basic/template project. I've also recorded a talk on this dapp and `basic_dao` is mentioned a lot, it would be confusing to rename it. Why not change the Motoko version, which isn't as prolific yet?

akhi3030:
> Also, why change this from `basic_dao` to `dao`? We should be explicit that this is a basic/template project. I've also recorded a talk on this dapp and `basic_dao` is mentioned a lot, it would be confusing to rename it. Why not change the Motoko version, which isn't as prolific yet?

I picked one to change at random.  Happy to rename the motoko version to `basic_dao`.  @chenyan-dfinity : are you able to work on that please?  I'll close this PR then.

PR #180, opened 2022-03-15
Encrypted Notes Motoko: replace HashMap with TrieMap

yvonneanne:
to get rid of warning about append performance


dfx-json:
hi @yvonneanne. thank you for contribution. as this PR is quite outdated, I would like to close it. if you would still like to include this example, I encourage you to create a new PR. thank you for your understanding.

PR #184, opened 2022-03-16
fix II refresh bug in svelte example

tim-dfn:
Fix II reload refresh bug


ililic (approved):
Verified on my dapp that the reload loop is fixed. LGTM

PR #185, opened 2022-03-17
Sa/hello/ux loader

wackyleo459:
Current Hello example dapp does not have any UI to indicate something is in progress, during the wait time for a response from a canister. 

Currently, when typing text into the input box and clicking the button, the time between invocation of the action and receipt of a response is long. This can lead users to feel like “nothing is happening” or that something is broken. 

**Implemented Solutions**
- When the “Click me” button is clicked, input is also disabled along with the button.
- Until response is rendered, a progress loader icon with a "Please Wait" message pops up.
- Once response is received, progress loader disappears. Input and button are re-enabled.

**Notes**
Same UI interface is applied to rust and motoko Hello example.

**Preview**
Loading
Screen Shot 2022-03-17 at 4 12 46 PM

 Finished
Screen Shot 2022-03-17 at 4 10 57 PM

### DIFF
--- a/motoko/hello/src/hello_assets/src/index.html
+++ b/motoko/hello/src/hello_assets/src/index.html
@@ -16,6 +16,23 @@
         <input id="name" alt="Name" type="text" />
         <button type="submit">Click Me!</button>
       </form>
+      <div id="loader">
+        Please Wait
+        <div class="lds-default">
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+          <div></div>
+        </div>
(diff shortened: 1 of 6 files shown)


motoko/hello/src/hello_assets/src/index.html
```diff
@@ -16,6 +16,23 @@
         <input id="name" alt="Name" type="text" />
         <button type="submit">Click Me!</button>
       </form>
+      <div id="loader">
+        Please Wait
+        <div class="lds-default">
+          <div></div>
```
jason-icp, line 22: why all the empty divs? could you possibly accomplish the intent here with css?

wackyleo459: those divs are all part of the individual dots making up the loader

dprats (approved):

PR #202, opened 2022-04-01
Dip721 NFT Example Motoko

jason-icp:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


sesi200 (approved):
I think this is ok to release as a first version, but we should still improve it to make sure the [minting tool](https://github.com/dfinity/experimental-minting-tool) works properly with this example.

Also, our documentation talks about the capability of serving certified assets when deployed to ic0.app, for which I think we should not just have a Rust example.

PR #206, opened 2022-04-14
Remove static mut from examples

adamspofford-dfinity:
Only one Rust example retained the legacy storage API and/or explicit `static mut`. This change removes it.


domwoe:
ah, haven't seen this before. This is implemented as part of #245

PR #238, opened 2022-05-09
[NET-959] Canister HTTP Requests Sample Dapp in rust.

qj-yu:
**Overview**
Creating a Sample Dapp for Canister HTTP Request feature. The sample dapp demonstrates how to make use of the feature to make off-chain HTTP calls, with example of pulling ICP <-> USDC exchange rate from [Coinbase API](https://docs.cloud.coinbase.com/exchange/reference/exchangerestapi_getproductcandles).

The Sample Dapp tries to avoid rate limiting factor of Binance API by pulling rates at a slow pace, and cache the rates that have been pulled locally in canister, for later user retrievals. 

The Sample Dapp makes use of Canister HTTP Request feature's `transform` function to normalize response received from Binance API, before passing response to Consensus. As such, some response headers are removed, such as: `Date`, `UUID`, `Via`, etc. which varies from request to request. 

**Requirements**
The Sample dapp is tested successfully on mainnet. 

**Considered Solutions**
Instead of slow pulling from Coinbase API, dynamic rate limit processing was considered. However, given rate limit is a common use case, instead of having canister developers building the solution repeatedly, the DFINITY team is considering building rate limit solution directly into the Canister HTTP Request feature within the IC.

**Recommended Solution**
The dynamic rate limit processing in IC will some extra time to complete. In the meantime, it's best to have canisters scatter out HTTP calls to the extent remote rate limiting will not be encountered.

**Considerations**
CICD pipeline setup is ideal as part of this merge, but not required.


rust/exchange_rate/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+ic-cdk = "0.5.0"
+ic-cdk-macros = "0.5.0"
+ic-types ="0.3.0"
+serde = "1.0.126"
+serde_derive = "1.0.126"
+cronjob = "0.3.17"
```
jplevyak, line 17: cronjob and tokio are not going to work.  getrandom and rand may not do what you want either depending on your requirements as they have no source of randomness, and indeed other than the call to management canister no way of getting any.  wasm-bindgen as well isn't going to work.

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub headers: Vec<HttpHeader>,
+    pub body: Vec<u8>,
+}
+
+type Timestamp = i64;
+
+thread_local! {
+    pub static FETCHED: RefCell<HashMap<Timestamp, Rate>>  = RefCell::new(HashMap::new());
+    pub static REQUESTED: RefCell<Queue<Timestamp>> = RefCell::new(Queue::new());
+    pub static SCHEDULER: JobScheduler = JobScheduler::new().unwrap();
```
jplevyak, line 49: You will have to build a replacement (or find, perhaps the NNS  folks have one) based on https://smartcontracts.org/docs/current/developer-docs/build/languages/motoko/heartbeats/

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+            return Ok(r);
+        }),
+        Err(m) => {
+            return Err(m);
+        }
+    }
+}
+
+#[allow(dead_code)]
+fn main() {}
```
jplevyak, line 251: See https://github.com/Psychedelic/DIP20/blob/9053e310f0e82d0251f786ae3f69a0d8961a2972/rust/token/src/main.rs#L629 for a good trick to see the candid.

jplevyak:
Generally looks good.  Not sure you need the scheduler at all since you could just update the rate on demand if it is stale.

rust/exchange_rate/src/lib.rs
qj-yu: Removed dependency to tokio_job_scheduler, and replaced with IC heartbeat method.

rust/exchange_rate/src/lib.rs
qj-yu: Adopted suggestion.

rust/exchange_rate/Cargo.toml
qj-yu: Removed cronjob, wasm-bindgen and rand dependencies.  Thanks for the suggestions!

qj-yu:
> Generally looks good. Not sure you need the scheduler at all since you could just update the rate on demand if it is stale.

Thanks for the review @jplevyak.  I have replaced tokio_job_scheduler dependency to IC heartbeat method to scatter out Binance API calls. The code is much simpler with this adjustments.  Thanks for guidance!

jplevyak:
Closed?

qj-yu:
Hmm, didn't intend to close at all.  Not sure whether I hit the wrong button. Reopening.

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub status: u64,
+    pub headers: Vec<HttpHeader>,
+    pub body: Vec<u8>,
+}
+
+thread_local! {
+    pub static FETCHED: RefCell<HashMap<Timestamp, Rate>>  = RefCell::new(HashMap::new());
+    pub static REQUESTED: RefCell<Queue<Timestamp>> = RefCell::new(Queue::new());
+
+    pub static RESPONSE_HEADERS_SANTIZATION: Vec<&'static str> = vec![
```
jplevyak, line 43: Doesn't need to be thread_local AFAIK

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+    });
+
+    // return rates for available ranges
+    return fetched;
+}
+
+/*
+Register the cron job which take the tip of the queue, and send a canister call to self.
```
jplevyak, line 112: s/take/takes/ s/send/sends/

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+Potentially, different nodes executing the canister will trigger different job during the same period.
+The idea is to gap the cron job with large enough time gap, so they won't trigger remove service side
```
jplevyak, line 114: ? remove service side ?  I am not sure what this refers to.   Perhaps just "rate limiting" ?

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+rate limiting.
+ */
+async fn get_next_rate() -> () {
+    let mut job_id: u64 = 0;
+
+    // Get the next downloading job
+    REQUESTED.with(|requested_lock| {
+        let mut requested = requested_lock.borrow_mut();
```
jplevyak, line 122: A RefCell is not a lock, so I would not call this _lock... _cell ?  And elsewhere.

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        if requested.size() == 0 {
+            return;
+        }
+
+        let job = requested.remove();
+
+        match job {
```
jplevyak, line 130: The only error returned by remove() is "queue is empty", so you can just do:

match requested_lock.borrow_mut().remove() { 
...
  Err(_) => return

  async fn get_next_rate() -> () {
      let mut job_id: u64 = 0;
  
      // Get the next downloading job
      REQUESTED.with(|requested_lock| {
          match requested_lock.borrow_mut().remove() {
              Ok(job) => {
                  // Job is a valid Job Id
                  FETCHED.with(|fetched_lock| {
                      match fetched_lock.borrow().get(&job) {
                          Some(_) => {
                              // If this job has already been downloaded. Only downloading it if doesn't already exist.
                              ic_cdk::println!(
                                  "Rate for {} is already downloaded. Skipping downloading again.",
                                  job
                              );
                              return;
                          }
                          None => {
                              // The requested time rate isn't found in map. Send a canister get_rate call to self
                              ic_cdk::println!("Fetching job {job} now.");
                              job_id = job;
                          }
                      }
                  });
              }
              Err(_) => return,
          }
      });
      get_rate(job_id).await;
  }

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                fetched.insert(requested, map.get(&requested).unwrap().clone());
+            } else {
+                // asynchoronously request downloads for unavailable ranges
+
+                // Simply putting the request to request queue. This queue will likely
+                // have duplicate entries, if users request same range of data multiple times.
+                // Double downloading is avoided right before the time of downloading by checking
+                // whether the data already exists in FETCHED map.
+                REQUESTED.with(|requested_lock| {
+                    let mut queue = requested_lock.borrow_mut();
```
jplevyak, line 91: Rust idiom is to reuse names, so this could be "let mut queue  = queue.borrow_mut();"

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                    match queue.add(requested) {
```
jplevyak, line 92: What prevents duplicates in the queue ?  Perhaps this should be a Set or MapSet ?

rust/exchange_rate/src/frontend/src/components/Chart.svelte
```diff
@@ -0,0 +1,251 @@
+<script>
+  // agent to connect to backend canister
+  import { backend_dummy } from "../../../declarations/backend_dummy/index.js";
```
qj-yu, line 3: @r-birkner could you please check in your local `backend_dummy/index.js`? We will replace it with `exchange_rate/index.js`.

r-birkner: You can create it yourself by running `dfx generate`. In the file, you will then have to replace the condition in line 19 with `true` to make it work for now.

rust/exchange_rate/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub value: String,
+}
+
+#[derive(Clone, Debug, PartialEq, CandidType, Eq, Hash, Serialize, Deserialize)]
+pub enum HttpMethod {
+    GET,
+}
+
+#[derive(CandidType, Deserialize, Debug)]
+pub struct CanisterHttpRequestArgs {
```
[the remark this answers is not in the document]
qj-yu: Might be useful to have it in `ic-cdk`. I'll reach out to SDK team.

rust/exchange_rate/src/lib.rs
qj-yu: Right, this static vector doesn't need to be.  Moved to global scope.

rust/exchange_rate/src/lib.rs
qj-yu: corrected.

rust/exchange_rate/src/lib.rs
qj-yu: *remote service side rate limiting :)  Fixed.

rust/exchange_rate/src/frontend/src/components/Chart.svelte
qj-yu: Thanks @r-birkner , that worked.

rust/exchange_rate/src/frontend/src/components/Chart.svelte
```diff
@@ -0,0 +1,250 @@
...
+
+  // loading spinner
+  import OverlayLoading from "svelte-overlay-loading";
+  import { Stretch } from "svelte-loading-spinners";
+
+	import chartjs from 'chart.js/auto'
```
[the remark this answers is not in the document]
qj-yu: Looks like this is fixed.

qj-yu:
Can we have a review on this merge request @dfinity/dx  and @dfinity/editorial teams?

rust/exchange_rate/README.md
```diff
@@ -0,0 +1,10 @@
+## Build the canister wasm file
+`cd rust/exchange_rate`
+`wasm-pack build --out-dir ./wasm/ --out-name exchange_rate.wasm --release`
+## Deploy the canister
+`dfx deploy`
+
+# Mainnet exchange_rate canister Id
+mfwph-oqaaa-aaaam-qafsq-cai
+
+Check the status of the canister: `dfx canister --network=ic status mfwph-oqaaa-aaaam-qafsq-cai`
```
sesi200, line 10: Could you please add some description what this example is all about?

rust/exchange_rate/canister_ids.json
```diff
@@ -0,0 +1,5 @@
+{
+  "exchange_rate": {
+    "ic": "464yd-nyaaa-aaaam-qag5a-cai"
```
sesi200, line 3: Should the mainnet deployment id be included in the example? Also, what's the difference between this id and the one mentionned in README (mfwph-oqaaa-aaaam-qafsq-cai)?

rust/exchange_rate/canister_ids.json
```diff
@@ -0,0 +1,5 @@
...
+  }
+}
```
sesi200, line 5: The canister exchange_rate_assets is not mentioned here. Is this on purpose?

rust/exchange_rate/src/main.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+/*
+Get rates for a time range defined by start time and end time. This function can be invoked
+as HTTP update call.
+*/
+#[update]
+async fn get_rates(range: TimeRange) -> RatesWithInterval {
+    // round down start time and end time to the minute (chop off seconds), to be checked in the hashmap
+
```
sesi200, line 125:
```suggestion
```

rust/exchange_rate/src/main.rs
```diff
@@ -0,0 +1,385 @@
...
+            return RatesWithInterval {
+                interval: i * REMOTE_FETCH_GRANULARITY as usize,
+                rates: fetched
+                    .into_iter()
+                    .filter(|r| r.0 % (i as u64) == 0)
+                    .collect(),
+            };
+        }
+    }
+    panic!("This shouldn't be happening! Couldn't find a inteval that can keep total data points count in {}.", MAX_DATA_PONTS_CANISTER_RESPONSE);
```
sesi200, line 177:
```suggestion
    panic!("This shouldn't be happening! Couldn't find an interval that can keep total data points count in {}.", MAX_DATA_PONTS_CANISTER_RESPONSE);
```

rust/exchange_rate/src/main.rs
```diff
@@ -0,0 +1,385 @@
...
+        set.insert(normalized_job);
+    });
+}
+
+/*
+Register the cron job which takes the tip of the job set, and sends a canister call to self.
+Potentially, different nodes executing the canister will trigger different job during the same period.
+The idea is to gap the cron job with large enough time gap, so they won't trigger remote service side
+rate limiting.
+ */
```
sesi200, line 197: I don't understand what cron thing this refers to. Maybe heartbeat?

rust/exchange_rate/src/main.rs
```diff
@@ -0,0 +1,385 @@
...
+    let mut should_fetch = false;
+    RATE_COUNTER.with(|counter| {
+        let state = counter.clone().into_inner();
+        if state == 0 {
+            should_fetch = true;
+        }
+        counter.replace((state + 1) % RATE_LIMIT_FACTOR);
+    });
+    if should_fetch {
+        get_next_rate().await;
```
sesi200, line 114: Awaiting a function within heartbeat is already a relatively risky thing, and this function can even make even more external calls. Can you add some comment about cycles use and/or canister safety here?

sesi200 (changes requested):
Thanks a lot for the hard work, @qj-yu. I think this will become one of the best and most-referred-to examples we have.

rust/exchange_rate/src/main.rs
qj-yu: adopted suggestion.

rust/exchange_rate/README.md
qj-yu: Yes, added purpose of the dapp, how to use, and canister behavior explanations.

rust/exchange_rate/canister_ids.json
qj-yu: You are right, this one is not needed. The canister_id referred in README.md is for mainnet, while this one, I believe is on some testnet.

rust/exchange_rate/canister_ids.json
qj-yu: This was because the frontend canister so far has only been tested locally, we haven't created it on mainnet (as the feature is not yet launched on mainnet).  Removed the canister_ids for mainnet for now.

qj-yu:
> Thanks a lot for the hard work, @qj-yu. I think this will become one of the best and most-referred-to examples we have.

Thanks so much for your review @sesi200 !

rust/exchange_rate/README.md
```diff
@@ -0,0 +1,64 @@
+## Purpose
+This canister is created to demonstrate usage of Internet Computer's newest feature: HTTP Requests
+for Canisters, where canister can make remote HTTP calls. Currently, the feature is limited to
+access secure (HTTPS) remote services that are served by nodes with IPv6 addresses. Trying to
+non-secure HTTP services or services with only IPv4 addresses will trigger Internet Computer errors.
```
sesi200, line 5:
```suggestion
for Canisters, where canister can make remote HTTP calls. Currently, the feature is limited to
access secure (HTTPS) remote services that are served by nodes with IPv6 addresses. Trying to access
non-secure HTTP services or services with only IPv4 addresses will trigger Internet Computer errors.
```

sesi200 (approved):
Really nice, thanks a lot! I have one more minor thing, feel free to merge when you think it's ready.

qj-yu:
Thanks for review and approval @sesi200 !   Merging now.

PR #241, opened 2022-05-16
Fix/tokens transfer

domwoe:
Tokens transfer example had a few issues
- Canister had the wrong name in `dfx.json`
- Cargo complained about `ic_types::Principal`
- Developers didn't know how to get `TOKENS_TRANSFER_ACCOUNT_ID_BYTES`
- Ledger canister is too big for application subnet (default since dfx v 0.10)
- Amount transferred to the tokens_transfer canister was not enough

@MarioDfinity would be nice if you could have a look.


roman-kashitsyn (approved):
LGTM

PR #250, opened 2022-06-03
fix defi_dapp/exchange.mo

mraszyk:
**Overview**
This PR fixes a bug in the Motoko implementation of the defi_dapp: the profix from executing a swap of tokens between two principals should be accounted to the dapp's principal rather than the ledger.

In addition, this PR bumps up the dfx version and dependencies.

### DIFF
--- a/motoko/defi/dfx.json
+++ b/motoko/defi/dfx.json
@@ -34,9 +34,12 @@
     "build": {
       "args": "",
       "packtool": ""
+    },
+    "replica": {
+      "subnet_type": "system"
     }
   },
-  "dfx": "0.9.3",
+  "dfx": "0.10.0",
   "networks": {
     "local": {
       "bind": "127.0.0.1:8000",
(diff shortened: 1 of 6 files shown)


motoko/defi/dfx.json
```diff
@@ -34,9 +34,12 @@
     "build": {
       "args": "",
       "packtool": ""
+    },
+    "replica": {
+      "subnet_type": "system"
     }
```
sesi200, line 40: Can we remove this? We shouldn't have examples that rely on the system subnet because almost nobody may deploy to those subnets.

sesi200 (changes requested):

sesi200:
Logic changes are approved by Tim

motoko/defi/dfx.json
mraszyk: It won't be possible to deploy the ledger canister because its WASM size exceeds the application subnet limits of 2MiB.

sesi200: AFAIK this only applies if `ic-cdk-optimizer` is not installed. Could you please test that and explain why this is here in the docs?

sesi200 (approved):

PR #271, opened 2022-06-14
fix: svelte motoko starter build

### LINKED ISSUE #249: svelte-motoko-starter: process is not defined
image

peterpeterparker:
**Overview**

Resolves #249

**Recommended Solution**

- inject `"undefined"` values as default values for the canister ID to replace `process.env` in any case - better render an application with hints to the user in case II or the backend canisters have not been deployed rather than not catching error "process is not defined"
- move rollup replacer as last script
- fix svelte build with pre-processor for svelte templates 
- generate source map only in development
- ditech `iife` for modern es modules bundle


domwoe:
@sesi200 Could you approve this please?

sesi200 (approved):

PR #273, opened 2022-06-20
Motoko dip-721-nft-container example : wrong array list order access ?

branciard:
**Overview**
I suspect a problem in the array access of this Motoko dip-721-nft-container example with ```List.get``` used.

My front-end was giving the wrong owner to a specific token id ( first of the list instead of the last ).

**Requirements**
Token ownership was fix after this change.

**Considered Solutions**
Retrieve the right item using explicit to token id match.

**Recommended Solution**
Retrieve the right item using explicit to token id match.

**Considerations**
I do not know the performance issue of this fix using ```List.find``` instead of .```List.get```

### DIFF
--- a/motoko/dip-721-nft-container/src/Main.mo
+++ b/motoko/dip-721-nft-container/src/Main.mo
@@ -31,7 +31,7 @@ shared actor class Dip721NFT(custodian: Principal, init : Types.Dip721NonFungibl
   };
 
   public query func ownerOfDip721(token_id: Types.TokenId) : async Types.OwnerResult {
-    let item = List.get(nfts, Nat64.toNat(token_id));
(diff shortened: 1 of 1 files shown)


sesi200 (approved):

sesi200:
@branciard Thank you very much for fixing this. Much appreciated!

Could you please look one more time over here: https://github.com/dfinity/cla/issues/35, it seems like a period is missing over there. Sorry for the trouble...

branciard:
CLA period done. Y're welcome @sesi200 . (Thank you for this very helpful [NFTs Workshop](https://www.youtube.com/watch?v=GLXCNl9Xpu4&list=PLuhDt1vhGcreT8T9Ox_h-rnY38ioUBh54&index=7]) Severin.)

PR #280, opened 2022-07-04
Basic bitcoin example (rust)

ielashi:
An example of how to use the Bitcoin and ECDSA apis to receive/send bitcoin from a canister.


rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did
```diff
@@ -0,0 +1,38 @@
...
+
+    "get_balance": (address: text) -> (satoshi);
+
+    "get_utxos": (bitcoin_address) -> (get_utxos_response);
+
+    "get_current_fee_percentiles": () -> (vec nat64);
+
+    "send": (record {
+      destination_address: text;
+      amount_in_satoshi: nat64;
```
THLO, line 36: Why not `amount: satoshi`?

rust/basic_bitcoin/src/basic_bitcoin/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+TARGET="wasm32-unknown-unknown"
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+pushd $SCRIPT_DIR
+
+# NOTE: On macOS a specific version of llvm-ar and clang need to be set here.
+# Otherwise the wasm compilation of rust-secp256k1 will fail.
+if [ "$(uname)" == "Darwin" ]; then
+  # On macs we need to use the brew versions
```
THLO, line 12: So, the documentation needs to point out that Mac users need `brew` installed, right?

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
```diff
@@ -0,0 +1,91 @@
+use ic_btc_types::{
+    GetBalanceRequest, GetCurrentFeePercentilesRequest, GetUtxosRequest, GetUtxosResponse,
+    MillisatoshiPerByte, Network, SendTransactionRequest,
+};
+use ic_cdk::{api::call::call_with_payment, export::Principal};
+
+// The fees for the various bitcoin endpoints.
```
THLO, line 7: Why are some called `_COST_CYCLES` and others `_FEE`?
I like the first option because it contains more information, and there is no risk that somebody confuses it for a Bitcoin fee.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
```diff
@@ -0,0 +1,91 @@
...
+        (GetUtxosRequest {
+            address,
+            network,
+            filter: None,
+        },),
+        GET_UTXOS_COST_CYCLES,
+    )
+    .await;
+
+    utxos_res.unwrap().0
```
THLO, line 51: We should note in the documentation that pagination is ignored here.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,259 @@
...
+            script_sig: Script::new(),
+        })
+        .collect();
+
+    let mut outputs = vec![TxOut {
+        script_pubkey: destination.script_pubkey(),
+        value: amount,
+    }];
+
+    let remaining_amount = total_spent - amount - fee;
```
THLO, line 160: Also here, shouldn't it be `fee*transaction_size / 1000`?

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
```diff
@@ -0,0 +1,53 @@
...
+
+const KEY_NAME: &str = "test";
+
+/// Returns the ECDSA public key of this canister at the given derivation path.
+pub async fn ecdsa_public_key(derivation_path: Vec<Vec<u8>>) -> Vec<u8> {
+    // TODO: Replace this principal with the management canister when it's available.
+    // For now, call a canister that provides a mock implementation.
+    let ecdsa_canister_id = Principal::from_text("ryjl3-tyaaa-aaaaa-aaaba-cai").unwrap();
+
+    // Retrieve the public key of this canister at derivation path [0]
```
THLO, line 12: Why `[0]`? The function handles arbitrary derivation paths, doesn't it?

rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did
```diff
@@ -0,0 +1,38 @@
...
+  utxos: vec utxo;
+  tip_block_hash: block_hash;
+  tip_height: nat32;
+  next_page: opt blob;
+};
+
+service : {
+    "get_p2pkh_address": () -> (bitcoin_address);
+
+    "get_balance": (address: text) -> (satoshi);
```
BenjaminLoison, line 28: Should use `bitcoin_address` instead of `text` here and for `send`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
```diff
@@ -0,0 +1,91 @@
...
+use ic_cdk::{api::call::call_with_payment, export::Principal};
+
+// The fees for the various bitcoin endpoints.
+const GET_BALANCE_COST_CYCLES: u64 = 100_000_000;
+const GET_UTXOS_COST_CYCLES: u64 = 100_000_000;
+const GET_CURRENT_FEE_PERCENTILES_FEE: u64 = 100_000_000;
+const SEND_TRANSACTION_BASE_FEE: u64 = 5_000_000_000;
+const SEND_TRANSACTION_FEE_PER_BYTE: u64 = 20_000_000;
+
+/// Returns the balance of the given bitcoin address.
```
BenjaminLoison, line 14: Bitcoin should take an uppercase here (cf https://bitcoin.org/en/vocabulary#bitcoin)

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
```diff
@@ -0,0 +1,91 @@
...
+///
+/// Relies on the `bitcoin_get_balance` endpoint.
+/// See https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance
+pub async fn get_balance(network: Network, address: String) -> u64 {
+    let balance_res: Result<(u64,), _> = call_with_payment(
```
BenjaminLoison, line 19: Should use `Satoshi` instead of `u64`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Sends a (signed) transaction to the bitcoin network.
+///
+/// Relies on the `bitcoin_send_transaction` endpoint.
+/// See https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_send_transaction
+pub async fn send_transaction(network: Network, transaction: Vec<u8>) {
+    let transaction_fee =
+        SEND_TRANSACTION_BASE_FEE + (transaction.len() as u64) * SEND_TRANSACTION_FEE_PER_BYTE;
+
+    let res: Result<(), _> = ic_cdk::api::call::call_with_payment(
```
BenjaminLoison, line 79: No need to precise `ic_cdk::api::call::` here.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,259 @@
...
+//!
+//! The wallet here showcases how bitcoin addresses can be be computed
+//! and how bitcoin transactions can be signed. It is missing several
+//! pieces that any production-grade wallet would have, including:
+//!
+//! * Support for address types that aren't P2PKH.
+//! * Caching spent UTXOs so that they are not reused in future transactions.
+//! * Option to set the fee.
+use crate::{bitcoin_api, ecdsa_api};
+use bitcoin::util::psbt::serialize::Serialize as _;
```
BenjaminLoison, line 11: What is ` as _` useful for ?

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+}
+
+/// Sends a transaction to the network that transfers the given amount to the
+/// given destination, where the source of the funds is the canister itself
+/// at the given derivation path.
+pub async fn send(
+    network: Network,
+    derivation_path: Vec<Vec<u8>>,
+    dst_address: String,
```
BenjaminLoison, line 57: Could use [`Address`](https://gitlab.com/dfinity-lab/public/ic/-/blob/06f262a40a83a9a5aaa76b444e2225a6754448a1/rs/bitcoin/types/public/src/lib.rs#L7).

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,259 @@
...
+            // We have enough inputs to cover the amount we want to spend.
+            break;
+        }
+    }
+
+    print(&format!("UTXOs to spend: {:?}", utxos_to_spend));
+    print(&format!(
+        "UTXO transaction id: {}",
+        Txid::from_hash(Hash::from_slice(&utxos_to_spend[0].outpoint.txid).unwrap()),
+    ));
```
BenjaminLoison, line 136: This looks a bit repetitive as the line just above.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
THLO: I was wondering the same thing.   :-)

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
```diff
@@ -0,0 +1,53 @@
...
+    let res: (ECDSAPublicKeyReply,) = call(
+        ecdsa_canister_id,
+        "ecdsa_public_key",
+        (ECDSAPublicKey {
+            canister_id: None,
+            derivation_path,
+            key_id: EcdsaKeyId {
+                curve: EcdsaCurve::Secp256k1,
+                name: String::from(KEY_NAME),
+            },
```
BenjaminLoison, line 23: Maybe having a small function returning the `key_id` used for both `ecdsa_public_key` and `sign_with_ecdsa` would simplify a bit the code.

rust/basic_bitcoin/src/basic_bitcoin/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// The bitcoin network to connect to.
+//
+// When developing locally this should be `Regtest`.
+// When deploying to the IC this should be `Testnet` or `Mainnet`.
+const NETWORK: Network = Network::Regtest;
+
+thread_local! {
+    // The derivation path to use for ECDSA secp256k1.
+    static DERIVATION_PATH: RefCell<Vec<Vec<u8>>>  = RefCell::new(vec![vec![0]]);
```
BenjaminLoison, line 18:
```suggestion
    static DERIVATION_PATH: RefCell<Vec<Vec<u8>>> = RefCell::new(vec![vec![0]]);
```

rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did
ielashi: It's subtle. The reason I wrote `amount_in_satoshi` is for clarity in the candid UI. If it's `amount: satoshi` the box will just say "amount (nat64)" and it wouldn't be clear what the unit is.

image

rust/basic_bitcoin/src/basic_bitcoin/build.sh
ielashi: Yes, exactly. We had this documentation as part of the dev preview, and we will have it on our portal documentation as well.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
ielashi: Agreed, copy pasta. Updated to use `_CYCLES` everywhere.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
ielashi: I appreciate the attention to detail, but let's not worry too much about casing, especially when it's an example and not a specification. Life's too short for that :)

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_api.rs
ielashi: Agreed, done.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
ielashi: I am using the types in the bitcoin crate only internally, but method signatures are using the (standard) address strings.

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
ielashi: Good catch. Updated.

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
ielashi: That part is subjective. I personally prefer in-lining when it's this small, since it's clearer to see everything and to copy for developers than having to look up another method.

rust/basic_bitcoin/src/basic_bitcoin/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+}
```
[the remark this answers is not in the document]
ielashi: Good catch, a `RefCell` isn't necessary here. Removed.

rust/basic_bitcoin/src/basic_bitcoin/src/types.rs
```diff
@@ -0,0 +1,62 @@
+use ic_cdk::export::{
+    candid::{CandidType, Deserialize},
+    serde::Serialize,
+    Principal,
+};
+
+#[derive(CandidType, Deserialize)]
+pub struct SendRequest {
+    pub destination_address: String,
+    pub amount_in_satoshi: u64,
```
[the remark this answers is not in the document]
ielashi: See above. In fact, the whole reason I defined `SendRequest` was for clarity in the Candid UI. Otherwise, I wouldn't have defined a `struct` at all and instead passed in both arguments directly to `send`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,259 @@
...
+    let destination = Address::from_str(&destination).expect("Invalid destination address.");
+
+    // Select which UTXOs to spend. For now, we naively spend the first available UTXOs,
+    // even if they were previously spent in a transaction.
+    let mut utxos_to_spend = vec![];
+    let mut total_spent = 0;
+    for utxo in utxos.into_iter().rev() {
+        total_spent += utxo.value;
+        utxos_to_spend.push(utxo);
+        if total_spent >= amount + fee {
```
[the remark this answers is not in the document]
ielashi: You're absolutely right. This is what happens when I try to work in a hurry. Will fix.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
ielashi: This is historical. At some point, this file was importing `Serialize` from bitcoin and `Serialize` from `serde`. The `as _` avoid the naming conflict, but it's no longer needed so I removed it.

rust/basic_bitcoin/src/basic_bitcoin/src/lib.rs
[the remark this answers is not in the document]
BenjaminLoison: I was initially meaning using `vec![]` instead of `vec![vec![0]]`.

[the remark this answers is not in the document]
ielashi: Good point. Changed to vec![]

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -204,20 +249,47 @@ async fn sign_transaction(
...
 
-fn sha256(data: Vec<u8>) -> Vec<u8> {
+fn sha256(data: &[u8]) -> Vec<u8> {
     let mut hasher = sha2::Sha256::new();
     hasher.update(data);
     hasher.finalize().to_vec()
 }
 
+// Converts a public key to a P2PKH address.
+fn public_key_to_p2pkh_address(network: Network, public_key: &[u8]) -> String {
```
BenjaminLoison, line 267: Just using [`Address::p2pkh` from the Bitcoin crate](https://docs.rs/bitcoin/0.28.1/bitcoin/util/address/struct.Address.html#method.p2pkh) may be simpler.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -100,26 +98,71 @@ pub async fn send(
...
+// Builds a transaction to send the given `amount` of satoshis to the
+// destination address.
+async fn build_transaction(
+    own_public_key: Vec<u8>,
+    own_address: &Address,
+    own_utxos: &[Utxo],
+    dst_address: &Address,
+    amount: u64,
+    fee_per_byte: MillisatoshiPerByte,
+) -> Result<Transaction, String> {
```
BenjaminLoison, line 110: It seems that the return type of this function could just be `Transaction`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// Builds a transaction to send the given `amount` of satoshis to the
+// destination address.
+async fn build_transaction(
+    own_public_key: &[u8],
+    own_address: &Address,
+    own_utxos: &[Utxo],
+    dst_address: &Address,
+    amount: u64,
```
BenjaminLoison, line 108: Could use `Satoshi` here.

BenjaminLoison (approved):
Looks good to me.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -204,20 +249,47 @@ async fn sign_transaction(
...
+
+    let mut full_address = data_with_prefix;
+    full_address.extend(checksum);
+
+    bs58::encode(full_address).into_string()
+}
+
+// A mock for rubber-stamping ECDSA signatures.
+async fn mock_signer(_derivation_path: Vec<Vec<u8>>, _message_hash: Vec<u8>) -> Vec<u8> {
+    vec![0; 64]
```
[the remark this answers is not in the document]
BenjaminLoison:
```suggestion
    vec![1; 64]
```

It seems that your answer:

> It's actually important. The signature from ECDSA is always 64 bytes. Converting to DER is taken care of in the sec1_to_der method, so the current logic takes that into account. No need for anything else here.

disappeared, I don't know if that was intentional.

Indeed the size varies when serializing in DER format.

What I am not happy with (but is easily solvable) is that we pass `vec![0; 64]` from `mock_signer` to `sec1_to_der` to know how many bytes the signed transaction will be. However in production we don't know if `sec1_signature[0] & 0x80 != 0` and `sec1_signature[32] & 0x80 != 0` will stand.
While for `vec![0; 64]` both conditions don't stand so we proceed to the `else` code blocks for both conditions and the problem is that these two blocks result in 2 less bytes (per input) than the `if` code blocks, so here in production the transaction size may be 2 more bytes long per input.

This detail looks important to me as a user providing the minimal fee per byte (1 Satoshi per byte IIRC) then because the signed transaction size is under-estimated then the transaction will be considered as invalid by the Bitcoin network.

The easy solution is to pass `vec![1; 64]` from `mock_signer` to `sec1_to_der` instead of `vec![0; 64]`, likewise both conditions will stand.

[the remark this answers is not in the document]
ielashi: Yes, exactly, that will make the DER signature to be the longest form. That implies that we'll sometimes be paying a slightly higher fee than we need to, but I think that's ok. The difference is very small.

rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did
```diff
@@ -0,0 +1,37 @@
...
+
+    "get_balance": (address: bitcoin_address) -> (satoshi);
+
+    "get_utxos": (bitcoin_address) -> (get_utxos_response);
+
+    "get_current_fee_percentiles": () -> (vec millisatoshi_per_byte);
+
+    "send": (record {
+      destination_address: bitcoin_address;
+      amount_in_satoshi: satoshi;
```
THLO, line 35: Maybe you answered this question and I missed it but wouldn't it be fine to just call it `amount` given that the unit is `satoshi`?
I'm not insisting on changing this but I'm just wondering if we gain much by calling it `amount_in_satoshi`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,325 @@
...
+    amount: u64,
+    fee_per_byte: MillisatoshiPerByte,
+) -> Result<Transaction, String> {
+    // We have a chicken-and-egg problem where we need to know the length
+    // of the transaction in order to compute its proper fee, but we need
+    // to know the proper fee in order to figure out the inputs needed for
+    // the transaction.
+    //
+    // We solve this problem iteratively. We start with a fee of zero, build
+    // and sign a transaction, see what it's size is, and then update then
```
THLO, line 117: The suggestion button is grayed out (?), so:
it's size -> its size
then update then update -> then update

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,325 @@
...
+        if (signed_tx_bytes_len * fee_per_byte) / 1000 == total_fee {
+            print(&format!("Transaction built with fee {}.", total_fee));
+            return Ok(transaction);
+        } else {
+            total_fee = (signed_tx_bytes_len * fee_per_byte) / 1000;
+        }
+    }
+}
+
+fn build_transaction_helper(
```
THLO, line 150: Given the large amount of "helper code", did you consider splitting the code into one module for the main functions and another module for the helper functions to keep things tidy?
I'm aware that this would require the reader to open up another file to examine a particular piece of source code but that might be preferable to having hundreds of lines of code in one module. What do you think?

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,325 @@
...
+        input: inputs,
+        output: outputs,
+        lock_time: 0,
+        version: 2,
+    })
+}
+
+// Sign a bitcoin transaction.
+//
+// IMPORTANT: This method is for demonstrational purposes only and it only
```
THLO, line 215:
```suggestion
// IMPORTANT: This method is for demonstration purposes only and it only
```

rust/basic_bitcoin/src/basic_bitcoin/src/types.rs
```diff
@@ -0,0 +1,62 @@
...
+    MalformedDestinationAddress,
+    MalformedSourceAddress,
+}
+
+#[derive(CandidType, Deserialize)]
+pub enum SignTransactionError {
+    InvalidPrivateKeyWif,
+    MalformedSourceAddress,
+    MalformedTransaction,
+}
```
BenjaminLoison, line 62:
```suggestion
```

These two types aren't used in this basic Bitcoin example.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
ielashi: I understand your concern. It is a bit subjective what's better in terms of readability (small modules over several files vs one larger module). I am personally in the latter camp, where I don't mind a file being a few hundred lines long, provided that it's coherent, especially that I find myself bikeshedding a lot on the naming of all the different files in the former case.

So, if it's ok with you, I'll keep things in one file, but I am renaming this method to not be called "helper", so now theoretically speaking it's not a helper method anymore :)

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
ielashi: I'd like to stick with the primitives. One benefit of that is that this code will match the logic written in Motoko.

THLO (approved):

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
```diff
@@ -0,0 +1,53 @@
+use crate::types::*;
+use ic_cdk::{call, export::Principal};
+
+const KEY_NAME: &str = "test";
```
BenjaminLoison, line 4: [As suggested @THLO](https://github.com/dfinity/examples/pull/281/files/aab31d9a4aecfcd7856113b6eb354926a14d2747#r916953714), we should update the key before we release the code. According to [this proposal](https://nns.ic0.app/#/proposal/68945), we should use `test_key_1` instead of `test`. After having tested, making this modification doesn't involve any problem when using the mock tECDSA canister locally. Do you confirm that `test_key_1` should be the `KEY_NAME` as you tested on the IC ?

rust/basic_bitcoin/src/basic_bitcoin/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+#[update]
+pub async fn get_current_fee_percentiles() -> Vec<MillisatoshiPerByte> {
+    bitcoin_api::get_current_fee_percentiles(NETWORK).await
+}
+
+/// Sends a (signed) transaction to the bitcoin network.
+#[update]
+pub async fn send_transaction(transaction: Vec<u8>) {
+    bitcoin_api::send_transaction(NETWORK, transaction).await
+}
```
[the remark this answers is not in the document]
ielashi: Good catch. Yes, leaving out of the `did` file is intentional. I'll remove this update call from the canister, because it's not practically useful.

rust/basic_bitcoin/src/basic_bitcoin/src/ecdsa_api.rs
ielashi: Yes, I confirm.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
```diff
@@ -0,0 +1,327 @@
...
+    let own_public_key = ecdsa_api::ecdsa_public_key(derivation_path.clone()).await;
+    let own_address = public_key_to_p2pkh_address(network, &own_public_key);
+
+    print("Fetching UTXOs...");
+    let own_utxos = bitcoin_api::get_utxos(network, own_address.clone())
+        .await
+        .utxos;
+
+    let own_address = Address::from_str(&own_address).unwrap();
+    let dst_address = Address::from_str(&dst_address).unwrap();
```
BenjaminLoison, line 64: These two lines should be at the beginning of the `send` function in order to avoid falling under this case:

> Unfortunately, this means that if your canister made a call and then panicked in the callback, the canister might never release the resources it allocated for the call.

[Source](https://mmapped.blog/posts/01-effective-rust-canisters.html#async)

rust/basic_bitcoin/dfx.json
```diff
@@ -0,0 +1,37 @@
...
+      "candid": "src/basic_bitcoin/basic_bitcoin.did",
+      "wasm": "target/wasm32-unknown-unknown/release/basic_bitcoin.wasm",
+      "build": "src/basic_bitcoin/build.sh"
+    },
+    "ecdsa": {
+      "type": "custom",
+      "package": "ecdsa",
+      "candid": "ecdsa.did",
+      "wasm": "ecdsa.wasm",
+      "build": ""
```
BenjaminLoison, line 17:
```suggestion
      "build": []
```

Makes the `Executing ''` goes away when running `dfx deploy`.

rust/basic_bitcoin/src/basic_bitcoin/src/bitcoin_wallet.rs
ielashi: In this case, there are no resources that we're allocating such that panicking after an `await` would cause trouble.

Panicking after an `await` is unavoidable in our case actually, since we don't know our `own_address` until we make a call to `ecdsa_public_key`, so moving these two lines to the top doesn't really change much.

PR #300, opened 2022-07-16
Send cycles when making http_request call

qj-yu:
**Overview**
Canister HTTP feature has started charging cycles. As a result, the `exchange_rate` sample dapp is updated to send cycles when making `http_request` call through IC.


sesi200 (approved):
Thank you very much, QJ!

PR #313, opened 2022-07-20
Testnet and mainnet deployment script for exchange_rate

qj-yu:
**Summary**

This adds a deployment script for exchange_rate sample dapp, to get it deployed to a testnet (or mainnet).


sesi200:
`dfx deploy` works for local deployments and the general public does not have any testnets for now. I'm not convinced this should be part of our public examples

sesi200:
We don't really have that for the examples repo. They usually  are very self-contained and ready to be copy/pasted everywhere. If this was setting up the same locally instead of on a testnet, then I'd be happy to have it here, but probably in a folder like `scripts`.

qj-yu:
Ok I have moved the deployment script to internal repo.   Closing this PR for now.

PR #317, opened 2022-07-21
fixed month-off-by-one error and removed circle on line chart

r-birkner:
Minor fixes in preparation to showcase the sample dapp.


qj-yu:
LGTM.  Thanks @r-birkner !

sesi200 (approved):

PR #327, opened 2022-08-09
update to dfx 0.11.1

sesi200:
updating to dfx 0.11.1
https://dfinity.atlassian.net/browse/SDK-622


rust/hello/webpack.config.js
```diff
@@ -86,14 +86,6 @@ module.exports = {
       template: path.join(__dirname, asset_entry),
       cache: false,
     }),
-    new CopyPlugin({
```
nathanosdev, line 89: Maybe we can also remove the `CopyPlugin` import on line 5 and uninstall the package using `npm uninstall --save copy-webpack-plugin`.

sesi200: good point. removed.

aisconnolly (approved):

PR #328, opened 2022-08-09
fix duplicate assets exception for motoko minimal counter dapp

nathanosdev:
Fixing issues with the Motoko minimal counter dapp caused by upgrading to dfx 0.11.1. Also made some minor code style fixes for consitency.


sesi200 (approved):
Thank you!

PR #329, opened 2022-08-09
fix duplicate assets exception for motoko phone book

nathanosdev:
Fixing issues with the Motoko phone book dapp caused by upgrading to dfx 0.11.1. Also made some minor documentation fixes for simplification.

I removed the generated declarations because we don't commit them in the other projects. I can go back and generate them for the other projects if we'd like to include them and be consistent.


sesi200 (approved):
Thank you!

PR #330, opened 2022-08-09
fix duplicate assets exception for motoko random maze dapp

nathanosdev:
Fixing issues with the Motoko random maze dapp caused by upgrading to dfx 0.11.1. Also made a small change to HTML to prevent negative numbers being used (which causes the BigInt constructor to throw an exception).


sesi200 (approved):
Thank you very much!

PR #332, opened 2022-08-12
Examples

StephanusFischer:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/motoko/cert-var/package.json
+++ b/motoko/cert-var/package.json
@@ -21,7 +21,6 @@
     "@dfinity/principal": "0.10.1",
     "assert": "2.0.0",
     "buffer": "6.0.3",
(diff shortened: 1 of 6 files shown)


sesi200:
Thank you very much for this, Sven! Since this is part of the update to dfx 0.11.1 I'd rather merge this into that branch. I created #333 as a replacement.

StephanusFischer:
Ok thanks Severin. Any other examples that you need help looking into?

sesi200:
Nothing urgent, but if you feel like experimenting with Svelte, you could take a look at `svelte-starter` or `svelte-motoko-starter` and see if you can get one (or both) of them under CI testing (see `.github/workflows` for examples of testing)

PR #334, opened 2022-08-18
Bump the bitcoin version to 0.28.1

THLO:
The PR bumps the bitcoin version to 0.28.1.
This version is better suited for follow-up development work.

In addition, there is a minor change: The version in the transaction is changed from 2 to 1, which is the commonly used version number.


sesi200 (approved):

ielashi:
@THLO Has this been tested end-to-end to ensure the example continues to run as expected?

PR #336, opened 2022-08-24
Fixing bug in mock signer

THLO:
Bug: It is possible that the current mock signer creates a mock signature that results in a transaction that is 1 byte smaller than the transactions signed with the proper ECDSA signing mechanism.
As a result, if we specify a low fee such as 1 sat/b, the current code sets the fee to `b-1` sat for a transaction of size `b` bytes, which causes the transaction to be rejected.

This PR fixes this problem by returning a mock signature with all bits set, both for the Rust and the Motoko version of the basic Bitcoin dapp.


sesi200 (approved):

ielashi (approved):
LGTM, thanks!

PR #338, opened 2022-08-25
rust/counter | reimplement motoko/counter in rust

JamesMcGuigan:
**Overview**
Counter is the simplest state management canister example, but was previously only implemented in Motoko.

This pull request ports the Motoko example to Rust, to highlight the major syntax differences between languages.

**Additions**
I have added in rust unit tests that can be run via `cargo test`

**Observations**
In Motoko, Nat is considered a first-class primitive, and state can be managed as a class variable:

In Rust `Nat32` / `Nat64` is a more convenient primitive, as `Nat` maps to `num-bigint::BigUInt` which is a heavyweight heap vector that doesn't implement the Copy trait (triggering various borrow checker puzzles).

Additionally, global state requires using RefCell indirection:
```
thread_local! { static COUNTER: RefCell<Nat> = RefCell::new(Nat::from(0)); }
```


rust/counter/README.md
```diff
@@ -0,0 +1,90 @@
...
+that represents the current value of the counter.
+
+By using the Motoko keyword `stable` when declaring the `counter` variable,
+the value of this variable will automatically be preserved whenever your canister code is
+upgraded. Without the `stable` keyword, a variable is deemed `flexible`, and its value
+is reinitialized on every canister upgrade, i.e. whenever new code is deployed to the canister.
+
+To learn more about these features of Motoko, see:
+* https://sdk.dfinity.org/docs/language-guide/motoko.html#_orthogonal_persistence
+* https://sdk.dfinity.org/docs/language-guide/upgrades.html#_declaring_stable_variables
```
sesi200, line 16: Old stuff from the Motoko original, should be adapted for Rust.

rust/counter/README.md
```diff
@@ -0,0 +1,90 @@
...
+*  `inc`, which increments the value of the counter; and
+
+*  `get`, which gets the value of the counter.
+
+## Prerequisites
+
+Verify the following before running this demo:
+
+*  You have downloaded and installed the [DFINITY Canister
+   SDK](https://sdk.dfinity.org).
```
sesi200, line 33:
```suggestion
   SDK](https://internetcomputer.org/docs/current/developer-docs/build/install-upgrade-remove).
```
SDK has moved

rust/counter/dfx.json
```diff
@@ -0,0 +1,23 @@
...
+      "type": "rust"
+    }
+  },
+  "defaults": {
+    "build": {
+      "args": "",
+      "packtool": ""
+    }
+  },
+  "dfx": "0.11.1",
```
sesi200, line 15:
```suggestion
```
Can we remove this line, please? With it removed, it will always run the currently installed version; with the line, as soon as a new version is released, this will cause issues.

sesi200 (changes requested):
Thank you very much, @JamesMcGuigan! I have a few nitpicks I'd like to see cleaned up, otherwise this is great.

JamesMcGuigan:
@sesi200 Thank you for your feedback and code review. I have implemented all your suggestions and rewritten the documentation to describe Rust. 

Could you please confirm that what I have written is technically correct, if there is any important information missing, and if the writing style is suitable for the examples repo.

Thank you.

rust/counter/README.md
JamesMcGuigan: I have rewritten this section in the README attempting to discuss the mechanics of Rust. Please can you confirm that what I have said is technically correct.

rust/counter/README.md
JamesMcGuigan: DONE + updated link in Motoko counter example

rust/counter/README.md
```diff
@@ -1,36 +1,49 @@
...
 
-By using the Motoko keyword `stable` when declaring the `counter` variable,
-the value of this variable will automatically be preserved whenever your canister code is
-upgraded. Without the `stable` keyword, a variable is deemed `flexible`, and its value
-is reinitialized on every canister upgrade, i.e. whenever new code is deployed to the canister.
+Canisters in Rust are single-threaded, thus it is safe (and simplest) to
+declare global state variable within a `thread_local!` block
+and wrapping `COUNTER` in a `RefCell` pointer.
+
+This is the equivalent of the `static` keyword in Mokoto, and the value of
```
sesi200, line 13:
```suggestion
This is the equivalent of the `stable` keyword in Mokoto, and the value of 
```
Typo?

sesi200 (changes requested):
Awesome, thank you very much! I think there is one typo left, but I want to confirm with you that I this really is a typo. Otherwise I'd like to merge this.

JamesMcGuigan:
Looking at the Motoko counter example, the correct keyword is indeed `stable` and not `static`
```
actor Counter {
  stable var counter = 0;
}
```
- https://github.com/dfinity/examples/blob/master/motoko/counter/src/Main.mo

Which agrees with the documention
- https://github.com/dfinity/motoko/blob/master/design/Stable.md

Thank you for spotting this. I have committed your code review update.

sesi200 (approved):

PR #352, opened 2022-09-13
DFX 0.12.0 update

dfx-json:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


sesi200:
Not ready to merge yet - we'll wait until 0.12.0 is out for real and then test with that version

.github/workflows/rust-nft-wallet-example.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Rust NFT Wallet Darwin
         run: |
           pushd rust/nft-wallet
-          dfx start --background
+          dfx start --background --host 127.0.0.1:8000
```
ericswanson-dfinity, line 21: Why setting the host (port) here?

motoko/defi/dfx.json
```diff
@@ -32,20 +32,15 @@
       "main": "src/DIP20/motoko/src/token.mo"
     }
   },
+  "networks": {
+    "local": {
+      "bind": "127.0.0.1:8000"
+    }
+  },
```
ericswanson-dfinity, line 39: Does this test work without a defined local network?

motoko/invoice-canister/install-local.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 dfx stop
-dfx start --background --clean
+dfx start --background --clean --host 127.0.0.1:8000
```
ericswanson-dfinity, line 6: Does this need to be forced to port 8000?

rust/basic_bitcoin/dfx.json
```diff
@@ -1,6 +1,6 @@
 {
   "version": 1,
-  "dfx": "0.11.1",
+  "dfx": "0.12.0-beta.2",
```
ericswanson-dfinity, line 3:
```suggestion
  "dfx": "0.12.0",
```

.github/workflows/rust-nft-wallet-example.yml
sesi200: Because of the following lines in `frontend/scripts/buildIndexHtml.js`:

```
const replacement = isNetworkLocal ? "localhost:8000" : "ic0.app";
main("ROLLUP_CONNECT_SRC_DOMAIN", replacement);
```

Not sure how to replace that

motoko/defi/dfx.json
sesi200: No, because it requires subnet type `system`, and I don't want to mess with networks.json on behalf of the user if they run the install script

motoko/invoice-canister/install-local.sh
sesi200: the e2e tests are hard-coded in multiple places to port 8000. There would probably be a way to move this somewhere else, but I didn't have the time to look into that

rust/basic_bitcoin/dfx.json
sesi200:
```suggestion
```
I think with 0.12.0 available we can even remove this entirely

.github/workflows/rust-nft-wallet-example.yml
sesi200: I tried to remove it, but I think something about svelte just refuses to work. Error messages are not helpful

motoko/invoice-canister/install-local.sh
sesi200: was able to remove it 😄

.github/workflows/rust-nft-wallet-example.yml
sesi200: Tracking in this ticket: https://dfinity.atlassian.net/browse/SDK-852

sesi200 (approved):
TODOs are in tickets, so merging this should be fine

Issue #292, opened 2022-07-12
Invoice Canister Cleanup Tasks


krpeacock:
- [x] Ensure that person without viewing permission cannot view an already verified invoice
  - invoice.mo line 283
- [x] Gitignore src/declarations
- [x] replace dfx stop with `#!/usr/bin/env bash;  set -e; dfx stop
- [ ] Add access control for creating new invoices
- [x] Check account for transfer fee during transfer
  - ICPLedger.mo line 179
- [x] pull magic numbers into constants
  - main.mo line 139
- [ ] Refactor permission checks to a method

AshtonEngberg:
Is it possible to uncheck " Check account for transfer fee during transfer " ? 

The current logic doesn't actually fix this problem--the way to resolve it is require a minimum payable amount for creating invoices (if the token is ICP or ICRC-1). Since an invoice requires at least one transfer by the invoice canister that the payable amount must cover (from invoice subaccount to principal subaccount), I'd suggest the minimum to be twice the transfer fee. 

Somewhat similar to this issue, https://github.com/dfinity/invoice-canister/issues/35, I'd like to fix both in future commits as part of the bounty since otherwise this current pull request is ready to go (uploading it soon).

AshtonEngberg:
More of a general question, but couldn't all the specific result return types anonymously declared in the function itself be moved into its own module (or extended as a part of the Types)? This would make the invoice canister's code much more concise, which presumably would make it easier to grok or iterate development unless there's a reason it should be styled the way it is?

PR #356, opened 2022-10-05
invoice canister cleanup

krpeacock:
**Overview**
Takes care of several cleanup tasks from #292


lwshang (approved):

domwoe:
Would be great if we could get this merged since we now have someone working on the bounty to work on the Invoice canister.

sesi200:
Need to fix tests first

domwoe:
@sesi200 Do you know what the issue is here?

sesi200:
Something in the node scripts about calling `rm` on a non-existent file is currently the error that happens, but I don't remember anymore

AshtonEngberg:
Does anyone know if deploying a ledger canister as per https://internetcomputer.org/docs/current/developer-docs/integrations/ledger/ledger-local-setup is going to be deprecated in favor of using `nns install`/`import`? 

This is where the error is resulting from (the `nns install` adds the ledger wasm and the did corresponding to the public version of the interface yet the test script is redeploying the ledger canister using the previous method which requires the private version of the did interface which doesn't explicitly exist afaik when using `nns install`).

sesi200:
Yes, it'll change to `nns install`, but that will still take considerable time. We want to make this the first plugins for dfx. I think it's fine to change to `nns install` for the moment, even if we'll have to change it around again later.

AshtonEngberg:
So it's possible to use `nns install` instead of manually downloading the ledger wasm and swapping its private and public did files while installing. Simply rerunning `nns install` and `nns import` after starting dfx --clean will overwrite the ledger canister already installed this way. One thing is though if using `nns-ledger` it has to run configured in the system type subnet, so it can't be (afaik) deployed other than using `nns install` which means all 7 of the other nns canisters will also be installed each time this command is run. 

Additionally, the first ident-1.pem file (in the icjs / nns repo) corresponds to one of the two test accounts the nns-ledger is created with which can replace using the MINTER identity. Unless there's a problem with resetting the icp ledger canister state this way for the E2E tests, this should work. I've got to go through the E2E testing to make sure the identities still correctly map, but I can post this as a fix to resolve this to finish merging this PR unless anyone sees a problem?

sesi200:
That would be great, @AshtonEngberg. Thank you for offering to help!

AshtonEngberg:
I finished the changes to get things working for the pull request. I forked examples, and pushed the changes as a new branch off this pr branch you can view here: https://github.com/AshtonEngberg/examples/tree/ashton/invoice-finish-cleanup-merge

In the base invoice/ there's a steps-to-complete-merge.txt which outlines what was done (not necessary but for thoroughness). 

While things work, using `dfx nns install` / nns-ledger might have drawbacks that are beyond my decision making. Using the nns-ledger as provided by `dfx nns install` means 1) the networks.json configuration has to match the specs for these canisters--is there a way to setup separate local network declarations? (not all projects require subnet type to be system, or the 8080 port). 2a) Resetting the nns-ledger afaik has to be done through rerunning `dfx nns install` (upgrade, install, are not available to run on it) which means all 8 nns canisters are deployed each time the install script is run. While not a problem in itself, is perhaps not optimal? 

Additionally is that the canister id import of the nns-ledger in the invoice.mo does not resolve correctly the first time the invoice canister is built (which must happen each time the reinstall script is run). Presumably from how the nns-ledger is installed, it's canister id is not produced the 'normal' way and does not show up in .dfx/canisterids.json. However, if the invoice canister is immediately built again it will resolve and compile correctly; this is equivalent to running deploy twice in a row. At this point, calls on the nns-ledger from the invoice canister do work as they normally would. This is noted in the new install script logic. 

If these are too much of an issue and do not have another means of resolution, reverting to the previous method of installing the ICP ledger canister could be used.

domwoe:
@AshtonEngberg  Thank you!

I think it is best to create a new PR.

AshtonEngberg:
Separate from the pull request for the invoice bounty?

domwoe:
> Separate from the pull request for the invoice bounty?

yep, I'd suggest we merge this first.

domwoe:
@krpeacock can we close this PR in favor of #429 ?

krpeacock:
@domwoe yes

PR #394, opened 2022-12-02
React Native

wackyleo459:
**Overview**
React Native starter application

This is a simple React Native IOS project that seeks to communicate with a backend canister (dapp) deployed on the Internet Computer. 

The emphasis is to introduce to other React Native developers an implementation of their RN app that can make update calls and queries to a deployed dapp.

**Note**
Current dependency`@dfinity/agent` is using a locally packed version of agent-js.
However, with the next released version agent-js (v0.14.2) this local dependency can be replaced.


wackyleo459:
The instructions to try out the example will take a little bit of time for installation (such as Xcode).
@dfx-json Hi Jason, I forgot to request a specific person for review. I'm not currently authorized to request that, so pinging you here. If you have the time to that would be great!

@peterpeterparker If you'd like to give this example a try and review too.

krpeacock (approved):
verified working on agent-js 0.15.0! I've removed the preview build and the `bls-verify` polyfill

PR #406, opened 2022-12-07
only run relevant workflows

sesi200:
**Overview**
As pointed out by @olaszakos most of our workflow runs are unnecessary. This PR makes it so that only relevant workflows get triggered, not all 90+.


olaszakos (approved):
Looks good

PR #411, opened 2022-12-08
fix outdated II installation instructions

sesi200:
**Overview**
As reported on [this forum post](https://forum.dfinity.org/t/error-when-deploying-svelte-examples-from-dfinity-samples-bank/16983/22), installation instructions for II are outdated in the examples.

### DIFF
--- a/motoko/defi/scripts/install.sh
+++ b/motoko/defi/scripts/install.sh
@@ -44,7 +44,7 @@ dfx canister call GoldenDIP20 setFee "(420)"
 
 ### === DEPLOY INTERNET IDENTITY =====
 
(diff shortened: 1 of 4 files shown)


peterpeterparker:
II can now be configured in `dfx.json` to be deployed locally automatically so that developer would have absolutely nothing to do. Instead of updating the doc maybe we can update the example to take avantage of this since the manual step can be useless?

sesi200:
Thanks for the tip, @peterpeterparker! I'd like to merge this as a quick fix, and I'll note down to review/update this when I find the time

peterpeterparker (approved):
Plan works for me @sesi200. Thanks!

PR #418, opened 2022-12-12
checkin declarations

tthebst:
`"prestart": "dfx generate exchange_rate"` Only works if canister already exists. If we check-in the declarations for both examples it is easier to deploy.


sesi200 (approved):
Nice, thanks!

PR #419, opened 2022-12-14
RUN-457: Add Canister Timers example

dfinity-berestovskyy:
Closes RUN-457


sesi200:
Could you please also add a small test so that it is easier to maintain the example? For example [like this](https://github.com/dfinity/examples/blob/master/.github/workflows/motoko-counter-example.yaml)

sesi200 (approved):

PR #420, opened 2022-12-16
IOS App with notifications and II integration

keplervital:
**Overview**
This example includes an IOS native application that serves a dapp, can authenticate using the internet identity and can receive push notifications. 

**Considerations**
1. To enable service workers the iOS app needs to have enabled support for bound domains ([WKAppBoundDomains](https://webkit.org/blog/10882/app-bound-domains/))
1. WebAuthn is only natively supported by web views if the domain that is making the credentials call would serve the file `/.well-known/apple-app-site-association` with the specified `appId`, and, the app has that domain set as an [associated domain](https://developer.apple.com/documentation/xcode/supporting-associated-domains), otherwise, it fails with a WebAuthn error for security reasons (it does not support a wildcard value)
    1. For this reason we've opted to use a new [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession) which opens a new window for the user and sends a callback to the native app when the user is successfully authenticated 
1. The iOS app supports notifications, however, to test this feature you'll need to have a physical iOS device since the xCode simulator is not able to register to get a device id

**Demo**

https://user-images.githubusercontent.com/119848388/208689656-02a3e25e-260d-4c00-b2b0-553afa0fdfea.mov


motoko/ios-notifications/dapp-demo/dfx.json
```diff
@@ -0,0 +1,34 @@
...
+  "canisters": {
+    "ios_notifications_api": {
+      "main": "src/api/main.mo",
+      "type": "motoko"
+    },
+    "internet_identity": {
+      "type": "custom",
+      "candid": "internet_identity.did",
+      "wasm": "internet_identity.wasm",
+      "build": "bash -c 'test -f internet_identity.wasm || curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_dev.wasm -o internet_identity.wasm; test -f internet_identity.did || curl -sSL https://raw.githubusercontent.com/dfinity/internet-identity/main/src/internet_identity/internet_identity.did -o internet_identity.did'",
```
frederikrothenberger, line 11: It's no longer necessary to download manually. You can use this: https://github.com/dfinity/internet-identity/blob/main/demos/using-dev-build/dfx.json

krpeacock:
Nice work! I'm excited to have this example

motoko/ios-notifications/dapp-demo/internet_identity.did
```diff
@@ -0,0 +1,205 @@
+type UserNumber = nat64;
```
frederikrothenberger, line 1: This file and the `wasm` should not be checked in, right?

motoko/ios-notifications/dapp-demo/src/www/src/pages/about.ts
```diff
@@ -0,0 +1,17 @@
+import { html, render } from "lit-html";
+import { Page, PageOptions } from "../types";
+
+const content = (homeUrl: string) => html`<div class="container">
```
frederikrothenberger, line 4: Does it make sense to have this page at all? I would either remove it or make it more useful (by linking back to the example in the github repo for example).

motoko/ios-notifications/README.md
```diff
@@ -0,0 +1,131 @@
...
+- [Readme: dapp-demo](dapp-demo/README.md)
+- [Readme: ios-dapp-demo](ios-dapp-demo/README.md) 
+
+## Disclaimer
+
+This is an example dapp that demonstrates the potential of integrating a dapp with native apps. 
+
+Please be mindful when considering this code for production and be mindful of:
+
+1. The integration with II in this example is using a universal links which is known to be safer than custom app schemes that are not bound to the appID. To enable this in your production code you should look at setting up [universal links](https://developer.apple.com/documentation/xcode/supporting-universal-links-in-your-app).
```
frederikrothenberger, line 130: Maybe a special section called `security considerations` would make sense? Given that _not_ using universal links is just insecure I would also word this prompt a bit stronger.

motoko/ios-notifications/dapp-demo/src/www/src/auth.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  public loginType(): AuthLoginType {
+    if (Auth.currentURL().searchParams.has(AuthLoginType.Mobile)) {
+      return AuthLoginType.Mobile;
+    }
+
+    return AuthLoginType.Desktop;
+  }
+
+  public static async create(options?: {
```
frederikrothenberger, line 63: This still creates the delegation keypair in the dapp by agent-js, right? I think it is fine this way for this example, but I would add an explanation somewhere that the private key could be kept in the secure enclave (generated by the native) app and the public key sent over to the web view (and the delegation back), which would be more secure an make the design resistant to XSS vulnerabilities in the dapp.

motoko/ios-notifications/dapp-demo/src/www/src/pages/about.ts
frederikrothenberger: Ah, I just realized this is used in the push notification show-case. Still, linking back to the examples repo would be nice. 😉

motoko/ios-notifications/dapp-demo/internet_identity.did
keplervital: true, will add to `.gitignore`

motoko/ios-notifications/README.md
keplervital: Great suggestion, i've just added a `security considerations` section to the readme

motoko/ios-notifications/dapp-demo/src/www/src/auth.ts
keplervital: Yes, it's still being created in the dapp. 

i've added the suggestion of generating the keypair from the native app in the security considerations section of the readme.

PR #423, opened 2022-12-23
Use fragment for delegation callback

keplervital:
**Overview**
Update example ios notifications app to use fragment for the delegation callback.


frederikrothenberger (approved):
Thanks!

PR #432, opened 2023-01-19
Change 'Mokoto' to 'Motoko' in example documentation. 

dskloetd:
Change 'Mokoto' to 'Motoko' in example documentation. 
I'm assuming that was a typo.


ggreif (approved):
Thanks for spotting these! Indeed, these are misspellings.

PR #437, opened 2023-01-20
chore: bumps agent-js to 0.15.1

krpeacock:


krpeacock:
Too much effort to refactor this for now

PR #457, opened 2023-02-09
File Upload Example in Rust

victorcastro89:
**Overview**

I have created an example of how to upload a file in chunks using rust

**Requirements**

make build
make install 
Access the front end canister and upload a file

**Considered Solutions**

**Recommended Solution**

**Considerations**


sesi200:
Hi @victorcastro89, thanks a lot for your PR! Before I dive deep into the details: This looks very similar to the [asset canister](https://github.com/dfinity/sdk/blob/master/src/canisters/frontend/ic-certified-assets/src/lib.rs), which is already coded in Rust (and I assume the Motoko file upload is based on that in turn). I feel like this is very similar to the asset canister and I don't quite see the benefit of adding basically the same to the examples repo as well, just with some features removed. The maintenance effort of an extra example is not trivial. Do you see a big benefit this example carries over just linking to the main asset canister repo?

dfx-json:
thank you for your contribution. as this PR is a bit outdated, I will be closing it. if you would still like to submit your changes, please do so in a new PR. thank you for your understanding.

PR #465, opened 2023-02-27
Completed Invoice Canister BNT-2 Bounty 

atengberg:
This is the completed work for the grant-rfps bounty [BNT-2: Invoice Canister](https://github.com/dfinity/grant-rfps/issues/2). 

The major change is support for ICRC1 token standard which is introduced through a type that allows for not only additional ICRC1 tokens but additional standards as well. Building on the motivation and design of the original invoice canister's implementation and keeping in mind the code is intended to be a part of the examples repository, there is extensive documentation and testing.

As there's a lot involved in the completion of this bounty, to make it easier to specifically review the bounty's requirements have been fulfilled, there's an included [bounty task checklist](https://github.com/atengberg/examples/blob/ashton/invoice-bnt2/motoko/invoice-canister/bounty-task-checklist.md) outlining the tasks (and their subtasks) each with a summary explanation and mapped to relevant code files/lines (this could be removed before the PR would actually be merged).


domwoe:
Thanks @atengberg! I'll have a look later today. @krpeacock It would be great if you could have a look as well, since you have much more experience with the invoice canister.

krpeacock:
This is a doozy of a review and I'm just getting started, but I'm so happy that you've done such a good job with this project!

motoko/invoice-canister/dfx.json
```diff
@@ -1,26 +1,48 @@
...
+      "type": "custom",
+      "wasm": ""
+    },
+    "icp_ledger_canister": {
+      "type": "custom",
+      "wasm": "src/token-ledger-canisters/icp/ledger.wasm",
+      "candid": "src/token-ledger-canisters/icp/ledger.public.did",
+      "build": ""
+    },
+    "icrc1_token_ledger_canister_ex1": {
```
krpeacock, line 29: Since we need a couple examples by default, I wonder if it would make sense to use `ckBTC` and `SNS-1`

motoko/invoice-canister/dfx.json
```diff
@@ -1,26 +1,48 @@
...
+      "remote": {
+        "id": {
+          "ic": "ryjl3-tyaaa-aaaaa-aaaba-cai",
+          "local": "ryjl3-tyaaa-aaaaa-aaaba-cai"
         }
+      },
+      "type": "custom",
+      "wasm": ""
+    },
+    "icp_ledger_canister": {
```
krpeacock, line 23: Why is `icp_ledger` different from `nns-ledger`?

motoko/invoice-canister/dfx.json
atengberg: Re: earlier comment: Thanks, it was a great opportunity to get to know core aspects of and developing for the IC. 
 
As far as ckBTC and SNS-1... at the time, the `dfx sns install` wasn't ready to be used along side `dfx nns install`, although it is what I originally wanted to do for an example of an ICRC1 token (using both the `dfx nns` and `dfx sns` commands in the main project). Additionally, because ckBTC/ckTESTBTC requires mainnet connection and local replicas can't connect to the mainnet (for now) , it didn't seem appropriate to add to them to the main project (although I do use the correct values for the ckTESTBTC token canister in the readme as an example of adding a supported token).

For both ckBTC and SNS-1 dedicated examples in the examples/ subdirectory could be more suitable to demonstrate aspects specific to either (particularly if at least mock ckBTC canisters could be deployed to make local replica development possible). This thought is finished in the next reply...

krpeacock: Your reasoning makes sense!

I think that any ICRC-1 canister could be deployed as a "mock" `ckBTC` canister for the purposes of local development, apart from the minting and liquidating specifics. Someone may want a more fully fledged mock for those services in their application, but I definitely agree that it's beyond the scope of this repo

Simply by using an ICRC-1 canister like you are currently doing locally, and using the `"remote" -> "id"` config for those canisters in `dfx.json` would probably be sufficient, if we wanted to go that direction

krpeacock:
Apart from the github workflow which will probably just need one or two more changes to go green, this is looking really good! I've tested local deployment of the app and the examples, ran the tests, and I'm overall very happy with the quality of documentation and the many refinements you've added here

motoko/invoice-canister/dfx.json
atengberg: This overlaps with the previous comment/reply. 

The answer is that when I started the work there was motivation to show how `dfx nns install` could be used to deploy a local instance of the ICP ledger canister. Integrating that led to a deployed ICP ledger whose did file conformed to the candid installed with `dfx nns import`. Coincidentally at the time I had also installed another ICP ledger with the Motoko based actor class to understand how to adapt the ICP ledger to the Invoice Canister's API. Since their dids are not entirely uniform this is what ultimately led to the `SupportedToken` variant because by that point I was figuring out how to also include the ICRC1 standard (which also had its own Motoko actor class reference added). 

After that was implemented, it seemed the best way to balance the complexity of including all those and demonstrating that the `SupportedToken` type works, was to only use `dfx nns install` along side the wasms/dids of the Rosetta repository, since less hard-coded Motoko code/files would be necessary and IIUC the Rosetta repository is (at least in part) intended for external accountability. 

Thus the two ICP ledger canisters references are different because one is for the included Rosetta wasm/did and the other for that of `dfx nns install`. Without all the background, this is explained in the beginning of [clean-startup.mjs](https://github.com/atengberg/examples/blob/ashton/invoice-bnt2/motoko/invoice-canister/clean-startup.mjs). I would have included those specifics in the readme as well, but it seemed like there was enough going on in the readme already. 

I do agree that having an example of invoice canister code that includes a "standard set" of tokens would be useful, although maybe that would only use mock classes to represent all the involved token ledger canisters. As for the main project, if it wouldn't be too much adding an SNS-1 token with `dfx sns install` seems appropriate; although a dedicated e2e/integrated example of using the invoice canister for a sns dapp also seems useful. In any case, since the scope of the project was for completing the bounty, I did not develop the `SupportedToken` type as much as I would have liked--in fact, while working on the `motoko-seller-client` example it became clear how much overlap there might be with an independent library module the invoice canister could take the form of, or for that matter the `SupportedToken` type itself. 

In any case, hopefully this code base can help move things along.

atengberg:
Great! Thanks for taking the time to review it, hope it wasn't too much. I'm not sure how the github workflow works, but if there is something I can do to help let me know. Also, the bounty checklist is probably not necessary to be a part of the PR (though it'd still be apart of the commit history in case it might be useful in the future) if I should delete that or one of the reviewers wants to before the PR would actually be merged. Until the merge actually happens, I'll be sure to keep the branch synced with dfinity/examples.

motoko/invoice-canister/dfx.json
krpeacock: I think we can remove the included wasm. I only was using that originally because `dfx nns install` didn't exist yet at the time

atengberg: The reason I added it back after integrating `dfx nns install` was to provide an applied demonstration of the `Local Ledger` instructions as well as have another token in showing that `SupportedToken` type works for multiple tokens (along with the E2E tests). Also while `dfx nns install` can be used to deploy an ICP ledger canister, it might not always be feasible for a project because it comes as part of all the nns canisters that also must be installed (they are omitted from `dfx.json` as they are not needed, but they are all installed in the local replica as well when `dfx nns install` is used). Additionally, it gives an opportunity to show how `zx` can be used to automate swapping the private and public dids of the ICP ledger when deploying it with the wasm. 

The motivation was to have the main project demonstrate the "main" ways an ICP (and ICRC1) ledger could be deployed while also "proving" the `SupportedToken` type can handle many entries. I had actually added both the Motoko references for the actor classes as well (for a total of 6 tokens), but that introduced some other issues, and having the wasms/dids of the Rosetta repository seemed to be the most relevant (along with `dfx nns install`) since they are intended (IUCC) for "external accountability" and designed for "plug n play" functionality. 

If it doesn't make it too complicated, adding `dfx sns install` would then complete this idea particularly now that `dfx sns install` is working. Adding it along side `dfx nns install` would definitely be additional work that I thought could be a future task along with other features such as using Candb to autoscale invoice record storage, implementing CertifiedData or encryption with threshold ECDSA, adding (a rudimentary) official "status" to invoices to do things like proper refunding designating the consolidation address other than the creator's subaccount, providing JSON output to inbound HTTP calls, or refactoring the invoice canister functionality as independent module library components that could also be used elsewhere. 

Another possibility is including a project in the examples/ subdirectory that could be tailored towards supporting a specific set of "mainnet ready" tokens (such as the mainnet ICP ledger with the ICRC1 standard as the base) that could be more readily used dropped into another project. 

Although it could be removed, I thought it'd be more helpful as an educational example to include it.

krpeacock:
> Great! Thanks for taking the time to review it, hope it wasn't too much. I'm not sure how the github workflow works, but if there is something I can do to help let me know. Also, the bounty checklist is probably not necessary to be a part of the PR (though it'd still be apart of the commit history in case it might be useful in the future) if I should delete that or one of the reviewers wants to before the PR would actually be merged. Until the merge actually happens, I'll be sure to keep the branch synced with dfinity/examples.

I'd leave the bounty checklist in! Maybe it can live in the `docs` folder, though.

As for the CI, you can click on the Details where there is a red X or visit the checks tab. In this case, https://github.com/dfinity/examples/actions/runs/4359939522/jobs/7663887717 shows that `zx` isn't installed by the time that the `deployForTesting` job is trying to run. I think that just means that `npm ci` needs to be run first in `.github/workflows/motoko-invoice-e2e.yaml`

atengberg:
The bounty check list is moved to docs, and I added `npm ci` back to the make file for the first step of the e2e script. 

This was how the branch was originally configured, but I had dropped it due while integrating `dfx nns install` for a reason that's no longer relevant, although it should be there as you pointed out. This should unblock the workflow as well.

atengberg:
Another note, it was not just a matter of adding `npm ci`, as the project uses a system wide networks.json file, and it's not forced write it in the startup script, so I added setting it in the yaml file.

krpeacock (approved):

sesi200 (approved):
> the project uses a system wide networks.json file configured for the 8080 port and replica subnet type system

That's totally fine, as long as it is properly documented in the readme, which it is already.

> I updated the e2e yaml to configure this accordingly, unless someone thinks it would be better as part of the startup script?

Adding it to the e2e yaml is my preferred way of solving it. I don't like it if a project modifies my networks.json without my consent.

> does the networks configuration need to be removed after the e2e job finishes?

No, GH runners start from a clean slate for every test. You can screw up the system as much as you want and leave it in a terrible state, it will get deleted afterwards anyways.

PR #468, opened 2023-02-28
Fix Encrypted Notes Rust and improve the README file

aterga:
Currently, the Rust canister panicks with `[Canister r7inp-6aaaa-aaaaa-aaabq-cai] Panicked at ‘already mutably borrowed: BorrowError', src/encrypted_notes_rust/src/lib.rs:149:77)`.

To mitigate this, I propose removing the `users_invariant` checks from the `user_count` function. My reasoning is this: (1) IMHO, it’s not possible to corrupt the canister state via the public API to violate this invariant right now, and (2) enabling this check at each call of `user_count` would require refactoring the code to avoid borrowing already borrowed data. I don’t think it’s worth implementing such a refactoring at this point.

Additionally, the PR adds instructions on how to run Rust unit tests and tidies up the Dockerfile.


nikolay-komarevskiy:
Tested the local deployment with `sh ./deploy_locally.sh` and afterwards executed Rust tests `docker exec container_id cargo test`. The app works on `http://localhost:3000/` and 2 Rust unit tests have succeeded.

nikolay-komarevskiy (approved):
Great! Tested local deployment with Rust canisters. Works!

PR #469, opened 2023-03-01
Update Motoko threshold-ecdsa example

fspreiss:
Update the Motoko threshold-ecdsa example as follows:

* Use hex encoding (instead of candid blobs) as input and output representation for all types except the message
* Take the message in sign in raw form and hash it directly in the canister (instead of outside)
* Adapt the test.sh-script accordingly
* Adapt the README accordingly


sesi200 (commented):
LGTM, but I see you requested @andreacerulli's review. I'll wait for that review before I merge

sesi200 (approved):

PR #471, opened 2023-03-04
Basic referral feature in motoko

chipchipzero:
**Overview**

This is the basic referral feature for [sustainations_dao](https://github.com/triipme/sustainations_dao)
Please check [readme file](https://github.com/triipme/dfinity-examples/blob/motoko/basic_referral/motoko/basic_referral/README.md) for more information.

**Requirements**

* Install vessel: https://github.com/dfinity/vessel
* To run the test scripts, you need to download [ic-repl](https://github.com/chenyan2002/ic-repl/releases).

**Considered Solutions**

**Recommended Solution**

**Considerations**


atengberg:
In case you'd like to use the other Aviate Labs libraries (SHA224, CRC32) to do ICP type address computations (which also requires the Hex and Binary) you could use this as an [example](https://github.com/atengberg/examples/blob/ashton/invoice-bnt2/motoko/invoice-canister/src/invoice/modules/supported-token/token-specific/icp/Adapter.mo) (the project has the correctly configured packaget-set and vessel dhall files as well). The other consideration is adding ICRC1 address types as the mainnet ICP ledger supports that as well.

sesi200:
> The APLv2 is compatible with GPLv3, means code under APLv2 can be merged with GPLv3 code. The stricter copyleft protections of GPLv3 would then force us to bring everything under GPLv3 and that's why we cannot allow PRs of code under GPLv3.

I'm closing this because of the license. If you're willing to re-submit under a different license, I'm happy to take on a fresh PR.

PR #490, opened 2023-05-24
Stefan/context update

stefanneamtu:
**Overview**
The previous example was not making use of the `context` field in the transform function, which was not demonstrating the proper way of using it.

**Solution**
The `transform` function has now been updated to demonstrate correct use of the `context` field by providing the indices  of the data that must be saved (bucket start time and closing price index).

### DIFF
--- a/rust/exchange_rate/src/main.rs
+++ b/rust/exchange_rate/src/main.rs
@@ -203,12 +203,15 @@ async fn get_rate(job: Timestamp) {
     ic_cdk::api::print(url.clone());
 
     ic_cdk::api::print(format!("Making IC http_request call {} now.", job));
+
+    let bucket_start_time_index = 0;
+    let closing_price_index = 4;
     let request = CanisterHttpRequestArgument {
         url: url,
         method: HttpMethod::GET,
         body: None,
         max_response_bytes: Some(MAX_RESPONSE_BYTES),
-        transform: Some(TransformContext::new(transform, vec![])),
+        transform: Some(TransformContext::new(transform, vec![bucket_start_time_index, closing_price_index])),
         headers: request_headers,
     };
     match http_request(request).await {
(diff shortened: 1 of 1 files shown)


rust/exchange_rate/src/main.rs
```diff
@@ -203,12 +203,15 @@ async fn get_rate(job: Timestamp) {
...
+
+    let bucket_start_time_index = 0;
+    let closing_price_index = 4;
     let request = CanisterHttpRequestArgument {
         url: url,
         method: HttpMethod::GET,
         body: None,
         max_response_bytes: Some(MAX_RESPONSE_BYTES),
-        transform: Some(TransformContext::new(transform, vec![])),
+        transform: Some(TransformContext::new(transform, vec![bucket_start_time_index, closing_price_index])),
```
rumenov, line 214: This works only because we use indexes that are fairly small numbers.

I am wondering if it won't be more explicit if we encode the indexes for example using "0 4".
And later we do something like

` let mut parts = s.split_whitespace().map(|s| s.parse::<i32>());
`

rust/exchange_rate/src/main.rs
```diff
@@ -40,6 +41,11 @@ pub const DATA_POINTS_PER_API: u64 = 200;
 // 10 (bytes per field) * 6 (fields per timestamp) * 200 (timestamps)
 pub const MAX_RESPONSE_BYTES: u64 = 10 * 6 * DATA_POINTS_PER_API;
 
+#[derive(Serialize, Deserialize)]
+struct Context {
+    indexes: Vec<u8>,
```
rumenov, line 46: sorry, I was not clear enough I meant something different

I think here the fields should be 

```
bucket_start_time_index: usize
closing_price_index: usize

```

rust/exchange_rate/src/main.rs
```diff
@@ -238,13 +250,15 @@ async fn get_rate(job: Timestamp) {
     }
 }
 
-fn keep_bucket_start_time_and_closing_price(body: &[u8]) -> Vec<u8> {
+fn keep_bucket_start_time_and_closing_price(body: &[u8], context: &Vec<u8>) -> Vec<u8> {
```
rumenov, line 253: usually the more convension and the more consistent way of passing the content is just to use a slice - &[u8]

rumenov (approved):

sesi200 (approved):

PR #506, opened 2023-06-08
Composite query demo

Dfinity-skaestle:
Composite query calls allow fast calls between canisters using non-replicated execution.

This demo serves as an example of how to use the feature.


Dfinity-skaestle:
Closing in favor of https://github.com/dfinity/examples/pull/587

PR #507, opened 2023-06-09
chore: Bump dfx version to 0.14.1

dsarlis:


sesi200:
Dfx `0.14.0` was pulled because of a bug in the asset canister that breaks SPAs. Please use `0.14.1` instead

PR #511, opened 2023-06-14
Fix and review outdated documentation URLs from smartcontracts.org

nabdullindfinity:
**Overview**
Fix broken URLs pointing to smartcontracts.org. 
New documentation pages from https://internetcomputer.org/docs/current/ are now linked instead.
For each broken URL or unreachable page, an archive copy was checked to see which current document fits that link the best.

**Requirements**
Links to old documentation pages should be replaced with the most appropriate current document.

**Considered Solutions**
N/A

**Recommended Solution**
N/A

**Considerations**
N/A

PR #563, opened 2023-07-03
Fix and review outdated documentation URLs from smartcontracts.org

nabdullindfinity:
**Overview**
(second attempt after [pull/511](https://github.com/dfinity/examples/pull/511)) 
Fix broken URLs pointing to smartcontracts.org.
New documentation pages from https://internetcomputer.org/docs/current/ are now linked instead.
For each broken URL or unreachable page, an archive copy was checked to see which current document fits that link the best.

**Requirements**
Links to old documentation pages should be replaced with the most appropriate current document.

**Considered Solutions**
N/A

**Recommended Solution**
N/A

**Considerations**
N/A

### DIFF
--- a/hosting/static-website/README.md
+++ b/hosting/static-website/README.md
@@ -22,7 +22,7 @@ The example project requires the following installed:
 - git
 - dfx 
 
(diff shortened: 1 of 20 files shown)


sesi200 (approved):
Thank you very much!

PR #593, opened 2023-07-21
Update Sveltekit starter package dependencies. Vite 4 Svelte 4

zavx0z:
The dependencies in the `package-lock.json` for the Sveltekit starter project have been updated. This includes an update to `"lockfileVersion": 3` and bumping various dependencies to their latest versions such as `"@dfinity/agent": "^0.18.1"`, `"@sveltejs/kit": "^1.20.4` etc. Additionally, a new dependency `"@playwright/test": "^1.28.1"` has been added. It appears that various development tools and libraries have been updated to their latest version for better performance, compatibility, and security.


svelte/sveltekit-starter/dfx.json
```diff
@@ -19,7 +19,7 @@
 			"packtool": ""
 		}
 	},
-	"dfx": "0.12.0",
+	"dfx": "0.14.3",
```
sesi200, line 22:
```suggestion
```
I'd say removing that line is even better so we don't have to update the version all the time

sesi200 (commented):
Does the project deploy properly for you with dfx 0.14.3? I only see a bunch of errors:

sesi200:
I still get the same errors

dfx-json:
thank you for your contribution. I will be closing this PR as it does not seem that we will be able to move forward. if you would still like to submit your changes, please do so in a new PR. thank you for your understanding.

PR #599, opened 2023-07-28
docs(composite_query): update broken links

keplervital:
Links were broken due to `.html` and because of that response verification was failing.


Dfinity-skaestle (approved):
Thanks!

PR #601, opened 2023-08-02
Use newest Internet Identity in encrypted notes dapps

fspreiss:
Updates the encrypted-notes-dapp and encrypted-notes-dapp-vetkd to use the newest Internet Identity dev build.

This fixes the problem that only a single Internet Identity anchor can be created. In particular, attempting to create a second anchor led to the canister trapping with "the combination of timestamp 0 and nonce 0 has already been used". Especially for encrypted-notes-dapp-vetkd this was suboptimal because it prevented to show/prove that two different users cannot see each other's notes.

The dev build has some extra features that make Internet Identity easier to use in a development setting (such as a predictable captcha, no requirement for WebAuthn, etc.). Note that the dev-build was already used before, however.


fspreiss:
@sesi200, CI fails with
> The Replica returned an error: code 5, message: "Wasm module of canister rrkah-fqaaa-aaaaa-aaaaq-cai is not valid: Failed to deserialize wasm module with Unknown opcode 194"

You once said that Rust started to put new WASM-instructions into the binaries, so I guess that is what causes this error. 
Do you know how we can fix this?

frederikrothenberger (approved):
Thanks for cleaning this up 👍

PR #604, opened 2023-08-11
Extend vetKD IBE demo with Internet Identity login

fspreiss:
Extends the vetKD IBE demos (both Rust and Motoko) with Internet Identity login and allows to IBE-encrypt for a particular principal.

Also updates `.github/workflows/provision-darwin.sh` to make the installed node-js version configurable via an environment variable (leaving the default unchanged). This is done so that the CI runs for the vetkd Rust and motoko examples for macOS/darwin can use a newer node-js version (here 19.8.1) because with the old version (14.15.4) webpack fails with the error message "Field 'browser' doesn't contain a valid alias configuration" for the `@dfinity/identity` dependency of the `@dfinity/auth-client`.

### DIFF
--- a/.github/workflows/motoko-vetkd-example.yml
+++ b/.github/workflows/motoko-vetkd-example.yml
@@ -20,6 +20,7 @@ jobs:
       - name: Provision Darwin
         env:
           DFX_VERSION: 0.14.2
+          NODE_VERSION: 14.21.3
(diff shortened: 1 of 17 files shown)


andreacerulli (approved):
Thanks @fspreiss, LGTM

fspreiss:
Thanks for merging master, @domwoe!

PR #610, opened 2023-08-27
Update ledger.public.did

keccak-256:
Updating the ledger candid file as ICP is ICRC-1 compatible


letmejustputthishere:
thank you for your PR. this example does not use any `ICRC1` methods and the tests are setup in a way that they rely on the `public.did` not to contain any initialisation arguments, thus i will close this PR

PR #613, opened 2023-09-08
Unity iOS II integraion with DeepLink

vincent-dfinity:
1. This pr touches a lot of files, but most of them are folder structure reorganizations. For example, `unity_android_deeplink` -> `unity_ii_deeplink` which we can put both supports for Android & iOS.
2. Support II integration with DeepLink for Unity iOS platform, please check the [demo video](https://drive.google.com/file/d/1n2_gLcgFaiXqhBF_b3ZJ4mwW6A33s7CE/view?usp=drive_link) for details.
3. Simply the plugin workflow, we can reuse the same code path for Android and iOS.


frederikrothenberger (approved):
Unfortunately, I did not have time to actually test it out. 😬 

But if this is all that is needed to make it work on iOS, that's great. 👍

PR #629, opened 2023-12-07
QR code example: add a feature to make the result transparent

ulan:
This adds a "make transparent" option to the UI of the QR code page and the corresponding handling in the backend.

There are also minor drive-by fixes in Makefile and README.


rust/qrcode/Makefile
```diff
@@ -1,37 +1,26 @@
...
-.PHONY: upgrade
-.SILENT: upgrade
-upgrade: build
-	dfx canister install --all --mode=upgrade
+deploy: node_modules
+	dfx deploy
 
 .PHONY: test
 .SILENT: test
-test: install
```
ulan, line 27: `install` doesn't work well for iterative development when the canisters need to upgraded instead of installed.

sesi200 (approved):

PR #640, opened 2024-01-11
chore: Bump dfx

frederikrothenberger:
`dfx` is severly outdated in this repo and seems to break on gzipped wasm modules.


frederikrothenberger:
Close because I just breaks way too many things.

PR #706, opened 2024-02-02
chore: update rust pub-sub

domwoe:


jennifertrin (approved):
Thank you for the update.

[Not required to merge this PR]:

We should consider automatically updating to the most up-to-date dependencies. It may cause issues with existing code though.

PR #708, opened 2024-02-09
chore: use dfinity/setup-dfx github action

ericswanson-dfinity:
**Overview**
This is to be compatible with dfxvm once it's released.

I also updated all of the workflows to use actions/checkout@v4.  This PR is split into two commits: the first updates to actions/checkout@v4, the second updates to use dfinity/setup-dfx.

**Requirements**
These tests should install dfx now, and after dfxvm is released.

**Considered Solutions**
Another option would have been to copy the path setup from https://github.com/dfinity/setup-dfx/blob/main/action.yml into the provision scripts.
```
# Linux
echo "$HOME/.local/share/dfx/bin" >> $GITHUB_PATH
# Macos
echo "$HOME/Library/Application Support/org.dfinity.dfx/bin" >> $GITHUB_PATH
```

**Recommended Solution**
There's a github action now, so it seems nice to use it.  Also the workflows used to use a couple different methods of overriding the dfx version.  Now there's just one:
```
      - uses: dfinity/setup-dfx@main
        with:
          dfx-version: 0.14.2
```

**Considerations**
This should reduce maintenance costs going forward, since https://github.com/dfinity/setup-dfx should be able to perform its part.

### DIFF
--- a/.github/workflows/hosting-photo-storage-example.yml
+++ b/.github/workflows/hosting-photo-storage-example.yml
@@ -16,7 +16,7 @@ jobs:
   hosting-photo-storage-darwin:
     runs-on: macos-12
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v4
       - name: Provision Darwin
         run: bash .github/workflows/provision-darwin.sh
       - name: Hosting Photo Storage Darwin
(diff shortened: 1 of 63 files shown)


ericswanson-dfinity:
Closed because this change would mean one of two things:
- every workflow that doesn't pin the dfx version would use the latest dfx version (no reproducibility)
- every workflow would have to pin to the latest dfx version, so using a newer dfx version would mean touching every workfloow

PR #709, opened 2024-02-09
chore: update test workflows to use dfxvm

ericswanson-dfinity:
**Overview**
Make it so the example workflows will still work after dfxvm is released.

There were a few tests that call `dfx upgrade`.  Since I guess the intent of their maintainers is to always use the latest version of dfx, I changed those to call `dfxvm update`.

Also removed motoko/encrypted-notes-dapp-vetkd/provision-darwin.sh, which was only used to pin the Node version.  The related test now uses dfx 0.16.1.

**Requirements**
The workflows need to work now, and they need to work after we cut over the install script at https://internetcomputer.org/install.sh to install dfxvm.

**Considered Solutions**
Another option would be to use the dfinity/setup-dfx action: https://github.com/dfinity/examples/pull/708

The reason I closed that PR is that it would require one of the following:
- require each workflow to specify a version (meaning upgrading to a newer version of dfx means touching every workflow)
- most workflows use the latest dfx (meaning no reproducibility, as by default tests would then use whatever is the latest dfx version)

An advantage of the approach is that the setup-dfx workflow works with today's dfx install script, and will also work with dfxvm's install script.

**Recommended Solution**
Recommend going ahead with this PR, so that future updates to use a newer dfx only have to touch the provision scripts.

**Considerations**
After dfxvm 1.0 is released (in a few weeks time), we need to update the provision scripts to fetch https://internetcomputer.org/install.sh


motoko/encrypted-notes-dapp-vetkd/provision-darwin.sh
lwshang, line 1: The is almost the same as `github/workflows/provision-darwin.sh`.
The only two differences are:
* Node version
* DFX_VERSION

Probably, we can get rid of this slightly adjusted version.
Let this example use the `.github/` script and add a `dfxvm install 0.12.0` step if we DO have to use this particular version.
Or maybe it's fine to use the latest version of dfx.

ericswanson-dfinity: Changing the versions involved is beyond the scope of this PR, but I can eliminate this script

dfx-json (approved):

PR #710, opened 2024-02-16
feat: Sample app for query stats feature

Dfinity-skaestle:
**Overview**
With query stats, developers gain insights into the use of query calls in their canister.
This has previously been difficult, since no state can be persisted when executing query calls.

This example shows how to use the canister status to gain access to query statistics.
It also shows how to use a timer to periodically query them and calculate rates for query metrics.


rust/query-stats/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+name = "query_stats"
+version = "1.0.0"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+candid = "0.10"
+ic-cdk = { git = "https://github.com/dfinity/cdk-rs", rev = "c01fc1741eb3fec2df0bb4e5c5ff54fa027a6091" }
```
sesi200, line 11: Do we have to pin a specific cdk commit? I'd prefer if we can use an official version

rust/query-stats/dfx.json
```diff
@@ -0,0 +1,16 @@
...
+      "type": "rust"
+    }
+  },
+  "defaults": {
+    "build": {
+      "args": "",
+      "packtool": ""
+    }
+  },
+  "version": 1
```
sesi200, line 15:
```suggestion
```

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
+# Query stats
+
+## Prerequisites
+This example requires an installation of:
+
+- [x] Install the [IC SDK](../developer-docs/setup/install/index.mdx) at version 0.16.1 or higher.
```
sesi200, line 6: Link is broken on GH

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+- [x] Download the following project files from GitHub: https://github.com/dfinity/examples/
```
sesi200, line 7: Can you either recommend cloning the repo or point to a specific folder?

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Begin by opening a terminal window.
+
+This is currently only for use from within DFINITY, as the replica shipped with `dfx` as well as nodes on 
+mainnet do not enable the query stats feature at the moment.
```
sesi200, line 12: If this is true this should not be in this repo

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Once enabled on mainnet, the following instructions will work without the `--network` argument.
+
+From within DFINITY, it can be ran as following (e.g. with a deployment to the `small_with_query_stats` Farm testnet)
+
+Variable `$BN_HOSTNAME` points at the hostname of a boundary node. For example, from field `['bn_aaaa_records']['url']` of the output of a Farm deployment
```
sesi200, line 18:
```suggestion
```

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Installing the canister
+
+```
+dfx canister create --network="$BN_HOSTNAME" query_stats
```
sesi200, line 23:
```suggestion
dfx canister create --network ic query_stats
```
Choose between mainnet or local and adapt this everywhere where `BN_HOSTNAME` shows up

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+We will need the canister ID that has been assigned to our canister later.
+
+```
+dfx build --network=$BN_HOSTNAME query_stats
+dfx canister install --network=$BN_HOSTNAME query_stats
+```
+
+## Issuing load
+
+We now want to generate queries at a certain rate. One way to achieve this is by using `ic-workload-generator`, which is part of the [IC repo](https://github.com/dfinity/ic/tree/master/rs/workload_generator).
```
sesi200, line 35: Are there binaries available for downloading somewhere?

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```
+ic-workload-generator $BN_HOSTNAME -r 10 -n 6000 --call-method "load" -m query --payload "4449444c0000" --no-status-check --canister-id=$CANISTER_ID
+```
+
+
+## Get query stats
+
+There is a delay until query stats show up in the canister status. Depending on the configuration, this 
+can reach from multiple minutes to an hour.
```
sesi200, line 46: How does this behave on local? Is it easily testable there?

rust/query-stats/README.md
```diff
@@ -0,0 +1,75 @@
+# Query stats
```
sesi200, line 1:
```suggestion
# Query stats

This example is about...
```

rust/query-stats/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+use ic_cdk::api::management_canister::main::QueryStats;
+
+use std::cell::RefCell;
+use std::time::Duration;
+
+const TIMER_PERIOD_IN_SECS: u64 = 5;
+const QUERY_STATS_PURGE_AFTER_SECS: u64 = 60 * 60 * 24; // 24h, has to be much larger than query stats epoch
+const QUERY_STATS_RATE_FOR_LAST_SECS: u64 = 60 * 60; // 1h, has to be larger than query stats epoch
+
+// Unfortunately, we need to redefine this type, since the members are not public.
```
sesi200, line 14:
```suggestion
// Unfortunately, we need to redefine this type, since the members are not public.
```
Can this be fixed somehow? E.g. by including it in some appropriate crate?

sesi200 (commented):
Can you also add some basic tests that at a minimum show that the example still compiles? You can draw inspiration from the other examples

rust/query-stats/README.md
Dfinity-skaestle: This is currently true, but the canary rollout should start soon.

rust/query-stats/README.md
Dfinity-skaestle: No.
The workload generator is not strictly needed for this, but it's probably the easiest tool to get started.

What would you recommend instead? Should I recommend to run `dfx` calls in a bash loop instead?

rust/query-stats/README.md
Dfinity-skaestle: Not currently, but as discussed offline, we should probably enable it in the `dfx` replica. I will talk to people to come up with a suggestion on how to do that.

sa-github-api:
Dear @Dfinity-skaestle,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/454) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

dfx-json:
cloned by https://github.com/dfinity/examples/pull/769

PR #722, opened 2024-02-29
🎨 rewrote example

letmejustputthishere:
**Overview**

The example was outdated.

**Requirements**

Update the dependencies and README.


rust/tokens_transfer/src/tokens_transfer_backend/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::hash::Hash;
+
+use ic_cdk_macros::*;
+use ic_ledger_types::{
+    AccountIdentifier, BlockIndex, Memo, Subaccount, Tokens, DEFAULT_SUBACCOUNT,
+    MAINNET_LEDGER_CANISTER_ID,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Hash)]
```
domwoe, line 11: Why do you need to derive `Hash` ?

rust/tokens_transfer/src/tokens_transfer_backend/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+        amount: args.amount,
+        fee: Tokens::from_e8s(10_000),
+        // The subaccount of the account identifier that will be used to withdraw tokens and send them
+        // to another account identifier. If set to None then the default subaccount will be used.
+        // See the [Ledger doc](https://internetcomputer.org/docs/current/developer-docs/integrations/ledger/#accounts).
+        from_subaccount: None,
+        to: AccountIdentifier::new(&args.to_principal, &to_subaccount),
+        created_at_time: None,
+    };
+    ic_ledger_types::transfer(MAINNET_LEDGER_CANISTER_ID, transfer_args)
```
domwoe, line 38: It seems that your updated version of the example doesn't support interacting with the local version of the ledger, and then it wouldn't make sense to setup the local ledger at all, or am I missing something?

rust/tokens_transfer/src/tokens_transfer_backend/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+        .await
+        .map_err(|e| format!("failed to call ledger: {:?}", e))?
+        .map_err(|e| format!("ledger transfer error {:?}", e))
+}
+
+// Enable Candid export (see https://internetcomputer.org/docs/current/developer-docs/backend/rust/generating-candid)
+ic_cdk::export_candid!();
```
domwoe, line 45: Couldn't this be confusing as you need the candid extractor tool and you need to run the necessary commands. I think you'd need to at least mention this in the readme.

domwoe (commented):
Thanks for taking this on to bring it up to date.

Ideally, I'd say the token transfer example would use ICRC-1 and ICRC-2 interface or what do you think?

letmejustputthishere:
> Ideally, I'd say the token transfer example would use ICRC-1 and ICRC-2 interface or what do you think?
i think it's good to have an example for interacting with the ICP ledger standard, but we can create another one to interact with ICRC1 ledgers.

rust/tokens_transfer/src/tokens_transfer_backend/src/lib.rs
letmejustputthishere: the local and mainnet canister IDs are the same, thus it's ok to use `MAINNET_LEDGER_CANISTER_ID` here

rust/tokens_transfer/src/tokens_transfer_backend/src/lib.rs
letmejustputthishere: the links to the docs that i provided explain the necessary steps. as the users are instructed to copy paste the did they don't need to know about this for the sake of the example anyways. but i can add a section to the readme.

domwoe:
> i think it's good to have an example for interacting with the ICP ledger standard, but we can create another one to interact with ICRC1 ledgers. i can also add it to this example. what do you prefer?

Makes sense. My preference would be to rename this example to `icp_transfer` and create a new `token_transfer` example based on ICRC-1/2. The readme of the first one should also mention that the ICP ledger also supports ICRC-1 and link to the `token_transfer` example. Please also give a quick heads up to @andreacerulli, since he's listed as the owner of this example.

domwoe:
> should we call this `icrc1_transfer` instead then?

Although this would be the most precise name, I would keep it at `token_transfer`, just because newcomers might have no idea what icrc1 is but might look for an example to create or interact with a token. Hence, I'd prefer meeting them at common terminology and then educating them about the standards.

domwoe (approved):

letmejustputthishere:
@andreacerulli @domwoe i renamed the example and added `token_transfer` examples in rust and motoko, can you review pls

andreacerulli (approved):
Thanks @letmejustputthishere! LGTM, I spotted a small typo

PR #724, opened 2024-03-04
docs: Update docs to use new 'Tag' UI feature on Dev Docs.

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/hosting/godot-html5-template/README.mdx
+++ b/hosting/godot-html5-template/README.mdx
@@ -2,7 +2,7 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 # Godot HTML5 sample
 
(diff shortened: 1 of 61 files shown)


hosting/godot-html5-template/README.mdx
```diff
@@ -1,5 +1,9 @@
+import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
+
 # Godot HTML5 sample
 
+<MarkdownChipRow labels={["Sample project"]} />
```
ulan, line 5: Isn't this label a bit redundant? The title already says that this is a sample (ditto for other pages).

PR #730, opened 2024-03-07
chore(FI-1222): Add symlink to fix rust-defi tests

mbjorkqvist:
Add symlink `rust/defi/src/frontend/declarations -> ../declarations` to fix the `rust-defi` tests.


mbjorkqvist:
For the `motoko-defi` tests, the `declarations` directory is placed in `motoko/defi/src/frontend`, whereas for the `rust-defi` tests, the directory ends up in `rust/defi/src`. Most of the files are the same for both tests (symlinks from a lot of files and directories in the `rust/defi` directory to `../../motoko/defi`), and I couldn't figure out the reason for this difference.

dfx-json (approved):

PR #732, opened 2024-03-08
fix: Revert back to .md and add links to GitHub code

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


dfx-json (approved):

dfx-json:
I've approved, but can we double check everything looks OK before merging?

jessiemongeon1:
> I've approved, but can we double check everything looks OK before merging?

Yep, I've double checked that all the READMEs display properly in the repo.

PR #769, opened 2024-03-20
chore: add query stats example bypassing CLA

dfx-json:
Clone of https://github.com/dfinity/examples/pull/710


motoko/query_stats/src/query_stats_backend/main.mo
```diff
@@ -0,0 +1,28 @@
+import Nat "mo:base/Nat";
+import Time "mo:base/Time";
+import Principal "mo:base/Principal";
+
+actor QueryStats {
+
+  let IC = actor "aaaaa-aa" : actor {
```
letmejustputthishere, line 7: you should be able to import the management canister like [this](https://github.com/letmejustputthishere/icrc7_launchpad/blob/f08286ea083e1246f6d8fde0b8a4c08f26aa12b9/backend/factory.mo#L1) without specifying it's interface manually

PR #864, opened 2024-05-07
feat: RUN-975: Add WebAssembly SIMD example

dfinity-berestovskyy:
**Overview**
This example showcases different approaches to utilizing the new SIMD instructions: Rust auto-vectorization and SIMD intrinsics for matrix multiplication, a core operation in Machine Learning and Artificial Intelligence applications. The example compares various SIMD optimization techniques and their potential speedups.

**Requirements**
To run the example, the latest `dfx` version `0.20.2-beta.0` or later is required with WebAssembly SIMD enabled replica.


rust/simd/mat_mat_mul.wat
ulan, line 1: Was this file added by accident?

rust/simd/mat_mat_mul/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+    let pa = a.as_ptr() as *const u8;
+    let pb = b.as_ptr() as *const u8;
+
+    for _ in 0..ITERATIONS {
+        let res = add_mat_mul_4x4_generic::<f32>(K, pa, pb);
+        std::hint::black_box(res);
+    }
+    // Assert the generic and SIMD matrix multiplication results are the same.
+    assert_eq!(
+        add_mat_mul_4x4_generic::<f32>(K, pa, pb),
```
ulan, line 157: Wdyt about additionally having the simplest possible "school" multiplication of matrices as the expected value for assertion? It is not obvious to me that add_mat_mul_4x4_generic correctly computes matrix multiplication because it uses the same 4x4 tiling as other algorithms. If there is a bug in the tiling, we wouldn't catch it.

By simplest matrix multiplication I mean: r[i][j] = sum(a[i][k] * b[k][j]) for all i, j, k.

rust/simd/mat_mat_mul/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+/// Returns the number of instructions used for a loop performing
+/// element-wise multiplication of `224x4` tiles of matrices `A` and `B`
+/// using loop auto-vectorization for generic Rust operators.
+///
+/// The following line enables auto-vectorization using WebAssembly SIMD instructions.
+#[target_feature(enable = "simd128")]
+#[ic_cdk_macros::query]
+fn auto_vectorization_f32() -> u64 {
+    let a = mats::M_4X4_F32.repeat(K / 4);
+    let b = mats::M_4X4_F32.repeat(K / 4);
```
ulan, line 147: I was expecting that the number of rows in matrix A matches the number of columns in matrix B in order for multiplication to work.

It seems that we are multiplying matrix A with a transposed matrix B (rows and columns swapped?). If so, then let's add a comment about it to avoid confusing readers?

rust/simd/mat_mat_mul/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+    let counter = ic_cdk::api::performance_counter(0);
+    ic_cdk::println!(
+        "Generic WebAssembly (f32): {:>12} Wasm instructions",
+        fmt(counter)
+    );
+    counter
+}
+
+/// Returns the number of instructions used for a loop performing
+/// element-wise multiplication of `224x4` tiles of matrices `A` and `B`
```
ulan, line 139: Nit: looks like we are multiplying a 4x224 (4 rows and 224 columns) matrix A with a 224x4 matrix B to get a 4x4 resulting matrix? Initially I thought that the resulting matrix will be 224x224 if matrix A is 224x4.

rust/simd/README.md
```diff
@@ -0,0 +1,215 @@
...
+## Further learning
+
+1. Have a look at the locally running dashboard. The URL is at the end of the `dfx start` command: `Dashboard: http://localhost/...`
+2. Check out `mat_mat_mul` canister Candid user interface. The URLs are at the end of the `dfx deploy` command: `mat_mat_mul: http://127.0.0.1/?canisterId=...`
+
+### Canister interface
+
+The `mat_mat_mul` canister provide the following interface:
+
+- `generic_f32` &mdash; returns the number of instructions used for a loop performing element-wise multiplication of `224x4` tiles of matrices `A` and `B` using generic Rust operators.
```
ulan, line 186: See my comments below about the 224x4 dimensions.

rust/simd/Makefile
```diff
@@ -0,0 +1,44 @@
...
+.PHONY: build
+.SILENT: build
+build:
+	dfx canister create --all
+	dfx build
+
+.PHONY: install
+.SILENT: install
+install: build
+	dfx canister install heartbeat --argument 1
```
ulan, line 13: I don't understand the heartbeat and timer parts here. Is this a copy from another project?

rust/simd/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Conclusion
+
+WebAssembly SIMD instructions unlock new possibilities for the Internet Computer,
+particularly in Machine Learning and Artificial Intelligence dApps. This example
+demonstrates potential 10x speedups for matrix multiplication with minimal effort
+using just Rust's loop auto-vectorization.
+
+As shown in Example 2, integer operations also benefit, although with a more modest
+"2x" speedup.
```
ulan, line 206: I think we can get higher speedup for integer using 8-bit integers and doing 16x16 tiles. Just fyi, nothing actionable here.

ulan (approved):
LGTM with comments. Thanks!

dfx-json:
@dfinity-berestovskyy this example should also have a corresponding workflow for automated tests. it can be as simple as just deploying the canister locally.

dfinity-berestovskyy:
Thanks for the review, folks.
I've addressed the comments, updated the CODEOWNERS and added the tests. The tests fail for now, as there is no `dfx` version with SIMD support released yet.

ghost (approved):

dfx-json:
LGTM. Thank you @dfinity-berestovskyy.

dfx-json:
Pending merge until dfx supports simd and tests pass

PR #877, opened 2024-05-29
Example with guards and asynchronous code for Rust canisters

gregorydemay:
Add an example for advanced Rust canister development showing the intricacies between:

1. the IC asynchronous message model;
2. its implementation with the Rust `ic_cdk`;

and their impact on guards, which are usually used to try to maintain some invariants. This is fairly specific to the Rust ic_cdk, since Motoko took another approach, where every async/await call will be converted to an inter-canister call (see [asynchronous functions](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/advanced-features/async-code/#asynchronous-functions))


rust/guards/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+                    .entry(item.clone())
+                    .and_modify(|v| *v = true);
+            });
+        });
+        // Note that this is the callback of the previous item. Similarly to the above method,
+        // the same 2 scenarios apply and whether the guard is effective or not depends on whether
+        // the future was polled until completion in a single message.
+        if item == panicking_item {
+            panic!("panicking callback!");
+        }
```
dsarlis, line 77: I'm wondering if this can be a bit confusing for readers since it happens before the `await` point. I suppose if the `panicking_item` is the first in the list, then the code is not very meaningful -- anything else besides the first item is meaningful.

rust/guards/tests/tests.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(CandidType, Debug, PartialEq, Eq)]
+pub enum FutureType {
+    TrueAsyncCall,
+    FalseAsyncCall,
+}
```
dsarlis, line 189: Can this not be imported instead of copying its definition?

gregorydemay: I tried but this doesn't seem top work with the `crate-type` being `cdylib` (see https://users.rust-lang.org/t/does-crate-type-has-effect-on-integration-test/76634).  A solution would be to put the type in a separate crate that can be both imported in the canister and as a dev-dependency but this seems overkill here.

rust/guards/src/lib.rs
gregorydemay: > I'm wondering if this can be a bit confusing for readers since it happens before the await point.

Initially this originated from a conversation I had with Thomas M. asking whether panicking at step i would impact the state of the elements processed during previous steps i-1 ...1,0. This is what the comment above L.72 tries to explain `// Note that this is the callback of the previous item.` Do you think this needs more explanation?.

> I suppose if the panicking_item is the first in the list, then the code is not very meaningful

Absolutely. We could add a check that this is not the case, but this might "muddy" the message in the sense that this would be an additional guard that's not really useful to carry the point of this example. So I would tend to leave as is, WDYT?

gregorydemay (commented):
Thanks @dsarlis for the feedback!

dfx-json (approved):
LGTM. thanks for the nice example and pocket IC provisioning script!

rust/guards/tests/tests.rs
dsarlis: I agree it's an overkill so we can keep things as they are.

rust/guards/src/lib.rs
dsarlis: We could enhance the comment to mention explicitly that if the panicking item is the first in the list, then it becomes kind of a no-op but any other element creates the same situation as the above method.

This is not a strong opinion though, so we could also leave things as they are.

PR #885, opened 2024-06-10
Use management canister API provided by ic-cdk

domwoe:
Updated the sample code to use the functions and types provided by ic-cdk.

### DIFF
--- a/rust/threshold-ecdsa/src/ecdsa_example_rust/src/lib.rs
+++ b/rust/threshold-ecdsa/src/ecdsa_example_rust/src/lib.rs
@@ -1,8 +1,11 @@
-use candid::{CandidType, Principal};
-use serde::{Deserialize, Serialize};
+use candid::CandidType;
+use ic_cdk::api::management_canister::ecdsa::{
+    ecdsa_public_key, sign_with_ecdsa, EcdsaCurve, EcdsaKeyId, EcdsaPublicKeyArgument,
+    SignWithEcdsaArgument,
+};
 use ic_cdk::{query, update};
+use serde::Serialize;
 use std::convert::TryFrom;
-use std::str::FromStr;
 
 #[derive(CandidType, Serialize, Debug)]
 struct PublicKeyReply {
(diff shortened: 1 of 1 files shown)


jessiemongeon1 (approved):

rust/threshold-ecdsa/src/ecdsa_example_rust/src/lib.rs
```diff
@@ -104,25 +63,19 @@ async fn verify(
     message: String,
     public_key_hex: String,
 ) -> Result<SignatureVerificationReply, String> {
-    let signature_bytes = hex::decode(&signature_hex).expect("failed to hex-decode signature");
-    let pubkey_bytes = hex::decode(&public_key_hex).expect("failed to hex-decode public key");
+    let signature_bytes = hex::decode(signature_hex).expect("failed to hex-decode signature");
```
fspreiss, line 66: Is there an advantage of moving the value into decode (i.e., of removing the `&`)?

rust/threshold-ecdsa/src/ecdsa_example_rust/src/lib.rs
```diff
@@ -19,82 +22,38 @@ struct SignatureVerificationReply {
...
 
-    let (response,): (SignWithECDSAReply,) = ic_cdk::api::call::call_with_payment(
-        mgmt_canister_id(),
-        "sign_with_ecdsa",
-        (request,),
-        25_000_000_000,
-    )
-    .await
-    .map_err(|e| format!("sign_with_ecdsa failed {}", e.1))?;
+    let (response,) = sign_with_ecdsa(request)
```
fspreiss, line 51: Somewhat out of scopehere, but `sign_with_ecdsa` handles the cycles cost under the hood, and it would be great if the [SIGN_WITH_ECDSA_FEE constant](https://docs.rs/ic-cdk/0.14.0/src/ic_cdk/api/management_canister/ecdsa/mod.rs.html#9) that is [used](https://docs.rs/ic-cdk/0.14.0/src/ic_cdk/api/management_canister/ecdsa/mod.rs.html#32) internally was public, so that developers can notice if the fee changes (cc @sesi200).

fspreiss (approved):
Thanks, @domwoe!

rust/threshold-ecdsa/src/ecdsa_example_rust/src/lib.rs
domwoe: It was suggested by clippy

PR #880, opened 2024-06-05
Make Crypto team own all encrypted-notes and threshold-ecdsa dapps

fspreiss:
Adapts the CODEOWERS file so that the Crypto team (rather than the Crypto division) owns all encrypted-notes (incl. encrypted-notes-vetkd) and threshold-ecdsa dapps, i.e., for both Rust and Motoko. So far, /rust/threshold-ecdsa/ was owned by the Consensus team, while /motoko/threshold-ecdsa/ was owned by the Crypto team.

PR #886, opened 2024-06-10
Make Crypto team own the Motoko threshold-ecdsa dapp

fspreiss:
Adapts the CODEOWERS file so that the Crypto team, rather than the Crypto division, owns the Motoko threshold-ecdsa example dapp. The idea was to do this already in https://github.com/dfinity/examples/pull/880, but that was missed.


dfx-json (approved):

fspreiss:
After merging I received a notification that the rust-nft-wallet-darwin workflow failed: https://github.com/dfinity/examples/actions/runs/9461303863

PR #887, opened 2024-06-12
Use Wasm SIMD in the image classification example

ulan:
The mainnet and dfx version 0.20.2-beta.0 support Wasm SIMD. Sonos Tract also supports Wasm SIMD on `master`.

This PR updates the Sonos Tract dependency to the version that supports Wasm SIMD.

Now image classification can run in a query. The example is modified to add an option for running as a query or update.

### DIFF
--- a/rust/image-classification/src/frontend/src/index.js
+++ b/rust/image-classification/src/frontend/src/index.js
@@ -21,7 +21,7 @@ async function classify(event) {
 
   try {
     const blob = await resize(img);
-    let result;;
+    let result;
     if (document.getElementById("replicated").checked) {
       result = await backend.classify(new Uint8Array(blob));
     } else {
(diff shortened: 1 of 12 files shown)


rust/image-classification/src/frontend/src/index.js
```diff
@@ -11,15 +11,22 @@ async function classify(event) {
...
   button.disabled = true;
   button.className = "clean-button invisible";
+  repl_option.className = "option invisible";
   message.innerText = "Computing...";
   loader.className = "loader";
 
   try {
     const blob = await resize(img);
-    const result = await backend.classify(new Uint8Array(blob));
+    let result;;
```
dfx-json, line 24:
```suggestion
    let result;
```

dfx-json (approved):

PR #921, opened 2024-06-21
fix cert var example

nathanosdev:


dfx-json (approved):
Thank you for updating the example @nathanosdev !

PR #928, opened 2024-07-03
chore: update canister_logs examples with replicated query cases

maksymar:
This PR adds more cases to the canister logging examples in Rust and Motoko.

Specifically it adds a case of recording logs in a query method called in replicated mode.
It also adds separate tests for each canister logging example case.

### DIFF
--- a/motoko/canister_logs/Makefile
+++ b/motoko/canister_logs/Makefile
@@ -21,19 +21,24 @@ upgrade: build
 .SILENT: test
 test: install
 	# Test print via update call.
-	dfx canister call CanisterLogs print 'print via update call'
+	dfx canister call CanisterLogs print 'print via update'
 	dfx canister logs CanisterLogs \
-		| grep 'print via update call' && echo 'PASS'
+		| grep 'print via update' && echo 'PASS'
 
 	# Test print via replicated query call.
-	dfx canister call CanisterLogs print_query 'print via replicated query'
+	dfx canister call --update CanisterLogs print_query 'print via replicated query'
 	dfx canister logs CanisterLogs \
 		| grep 'print via replicated query' && echo 'PASS'
 
+	# Test print via non-replicated query call should NOT record the message.
+	dfx canister call --query CanisterLogs print_query 'print via non-replicated query'
+	! dfx canister logs CanisterLogs \
+		| grep 'print via non-replicated query' && echo 'PASS'
+
 	# Test trap, ignore failed call output.
-	-dfx canister call CanisterLogs trap 'trap_message'
+	-dfx canister call CanisterLogs trap 'trap via update'
 	dfx canister logs CanisterLogs \
-		| grep 'trap_message' && echo 'PASS'
+		| grep 'trap via update' && echo 'PASS'
 
 	# Test memory_oob, ignore failed call output.
 	-dfx canister call CanisterLogs memory_oob
--- a/motoko/canister_logs/Makefile
+++ b/motoko/canister_logs/Makefile
@@ -20,10 +20,15 @@ upgrade: build
 .PHONY: test
 .SILENT: test
 test: install
-	# Test print.
-	dfx canister call CanisterLogs print 'print_message'
+	# Test print via update call.
+	dfx canister call CanisterLogs print 'print via update call'
 	dfx canister logs CanisterLogs \
-		| grep 'print_message' && echo 'PASS'
+		| grep 'print via update call' && echo 'PASS'
+
+	# Test print via replicated query call.
+	dfx canister call CanisterLogs print_query 'print via replicated query'
+	dfx canister logs CanisterLogs \
+		| grep 'print via replicated query' && echo 'PASS'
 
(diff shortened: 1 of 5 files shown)


motoko/canister_logs/Makefile
```diff
@@ -20,9 +20,35 @@ upgrade: build
...
+	dfx canister logs CanisterLogs \
+		| grep 'print via replicated query' && echo 'PASS'
+
+	# Test print via non-replicated query call should NOT record the message.
+	dfx canister call --query CanisterLogs print_query 'print via non-replicated query'
+	! dfx canister logs CanisterLogs \
+		| grep 'print via non-replicated query' && echo 'PASS'
+
+	# Test trap, ignore failed call output.
+	-dfx canister call CanisterLogs trap 'trap via update'
```
mraszyk, line 39: why `-dfx`, also below?

maksymar: because trapped call fails which makes the `make test` to fail as well. but in this case this failure is expected, we need this call to trap and then see that the trap message appears in logs.

motoko/canister_logs/Makefile
```diff
@@ -20,9 +20,35 @@ upgrade: build
...
+	dfx canister logs CanisterLogs \
+		| grep 'trap via update' && echo 'PASS'
+
+	# Test memory_oob, ignore failed call output.
+	-dfx canister call CanisterLogs memory_oob
+	dfx canister logs CanisterLogs \
+		| grep 'StableMemory range out of bounds' && echo 'PASS'
+
+	# Test timer trap, assume it's been 5 seconds since the start.
+	sleep 2
```
mraszyk, line 49: Why sleep 2 and assuming that it's 5 seconds since the start?

motoko/canister_logs/Makefile
```diff
@@ -20,9 +20,35 @@ upgrade: build
...
+	dfx canister call --update CanisterLogs print_query 'print via replicated query'
+	dfx canister logs CanisterLogs \
+		| grep 'print via replicated query' && echo 'PASS'
+
+	# Test print via non-replicated query call should NOT record the message.
+	dfx canister call --query CanisterLogs print_query 'print via non-replicated query'
+	! dfx canister logs CanisterLogs \
+		| grep 'print via non-replicated query' && echo 'PASS'
+
+	# Test trap, ignore failed call output.
```
mraszyk, line 38: > ignore failed call output

what does this comment mean?

rust/canister_logs/Makefile
```diff
@@ -20,9 +20,44 @@ upgrade: build
...
+	-dfx canister call canister_logs trap 'trap via update'
+	dfx canister logs canister_logs \
+		| grep 'trap via update' && echo 'PASS'
+
+	# Test panic, ignore failed call output.
+	-dfx canister call canister_logs panic 'panic via update'
+	dfx canister logs canister_logs \
+		| grep 'panic via update' && echo 'PASS'
+
+	# # Test memory_oob, ignore failed call output.
```
mraszyk, line 48:
```suggestion
	# Test memory_oob, ignore failed call output.
```

motoko/canister_logs/Makefile
maksymar: I have update those comments with `-dfx`, PTAL.

motoko/canister_logs/Makefile
mraszyk: But I cannot just run
```
$ -dfx
-dfx: command not found
```
so how does that `-dfx` work?

motoko/canister_logs/Makefile
mraszyk: But why do we need any timing assumptions at all to fetch the logs?

motoko/canister_logs/Makefile
mraszyk: Could you please replace "failed call output" by "failed dfx command output" if this is what you mean?

rust/canister_logs/Makefile
maksymar: Done

motoko/canister_logs/Makefile
maksymar: > To ignore errors in a recipe line, write a ‘-’ at the beginning of the line’s text (after the initial tab). The ‘-’ is discarded before the line is passed to the shell for execution.

it's a GNU make synthax, see [here](https://www.gnu.org/software/make/manual/make.html#Errors)

maksymar: @mraszyk sorry, I merged the PR before addressing this comment. 

here's another cleanup PR, please take a look.

https://github.com/dfinity/examples/pull/929

motoko/canister_logs/Makefile
maksymar: Done here: https://github.com/dfinity/examples/pull/929/files#diff-2dc8abd0614871d2c2ed078f6d1aff60e18d5ee57e292ea61b79b007316be7e7R39

PR #929, opened 2024-07-03
chore: update canister logging examples with trapped query calls

maksymar:
This PR updates canister logging examples with trapped query calls in replicated and non-replicated modes.

It also improves tests' comments.

### DIFF
--- a/motoko/canister_logs/Makefile
+++ b/motoko/canister_logs/Makefile
@@ -41,6 +41,18 @@ test: install
 	dfx canister logs CanisterLogs \
 		| grep 'trap via update' && echo 'PASS'
 
+	# Test trap via replicated query call.
+	# Ignore failed dfx command output (so the test continues) and check the logs to contain the message.
+	- dfx canister call --update CanisterLogs trap_query 'trap via replicated query'
+	dfx canister logs CanisterLogs \
+		| grep 'trap via replicated query' && echo 'PASS'
+
+	# Test trap via non-replicated query call should NOT record the message.
+	# Ignore failed dfx command output (so the test continues) and check the logs to contain the message.
+	- dfx canister call --query CanisterLogs trap_query 'trap via non-replicated query'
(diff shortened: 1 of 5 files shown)


motoko/canister_logs/Makefile
```diff
@@ -36,19 +36,21 @@ test: install
...
+	- dfx canister call CanisterLogs memory_oob
 	dfx canister logs CanisterLogs \
 		| grep 'StableMemory range out of bounds' && echo 'PASS'
 
-	# Test timer trap, assume it's been 5 seconds since the start.
-	sleep 2
+	# Test timer trap.
+	# The timer is setup to trap every 5 seconds, so this test has to be called 
+	# at least 5 seconds after the start to record the trap log message.
+	sleep 2  # This delay considers previous tests execution time.
```
mraszyk, line 53: Why are you sure that the previous tests took at least three seconds?

motoko/canister_logs/Makefile
```diff
@@ -36,19 +36,21 @@ test: install
 		| grep 'print via non-replicated query' && echo 'PASS'
 
 	# Test trapped call is recorded in logs. 
-	# Ignore failed call output (so the test continues) and check the logs to contain the message.
-	-dfx canister call CanisterLogs trap 'trap via update'
+	# Ignore failed dfx command output (so the test continues) and check the logs to contain the message.
+	- dfx canister call CanisterLogs trap 'trap via update'
```
mraszyk, line 40: For my understanding: where does this `- dfx` syntax come from? I cannot use it in a bash shell.

motoko/canister_logs/Makefile
mraszyk: Wouldn't it be better to have a one-off timer with no delay?

dfx-json (approved):

mraszyk (approved):

PR #934, opened 2024-07-10
CRP-2532 remove mock API from Schnorr example since dfx now supports Schnorr

altkdf:
The rust tests were disabled because PocketIC doesn't yet support Schnorr but will in the future. Having this example without a mock seems more important than having tests enabled for a few weeks, therefore, the tests are disabled until PocketIC supports Schnorr.

### DIFF
--- a/.github/workflows/rust-threshold-schnorr-example.yml
+++ b/.github/workflows/rust-threshold-schnorr-example.yml
@@ -25,6 +25,7 @@ jobs:
         run: bash .github/workflows/provision-pocket-ic-server.sh
       - name: Rust Threshold Schnorr Darwin
         run: |
+          dfxvm default 0.22.0-beta.0
           dfx start --background
           pushd rust/threshold-schnorr
           make deploy
@@ -41,6 +42,7 @@ jobs:
         run: bash .github/workflows/provision-pocket-ic-server.sh
       - name: Rust Threshold Schnorr Linux
         run: |
+          dfxvm default 0.22.0-beta.0
           dfx start --background
           pushd rust/threshold-schnorr
           make deploy
--- a/.github/workflows/rust-threshold-schnorr-example.yml
+++ b/.github/workflows/rust-threshold-schnorr-example.yml
@@ -27,9 +27,10 @@ jobs:
         run: |
           dfx start --background
           pushd rust/threshold-schnorr
-          make mock
-          sleep 10
-          make test
+          make deploy
+# This should be re-enabled when Pocket IC supports threshold Schnorr
+# sleep 10
+# make test
   rust-threshold-schnorr-linux:
     runs-on: ubuntu-20.04
     steps:
@@ -42,6 +43,7 @@ jobs:
         run: |
           dfx start --background
           pushd rust/threshold-schnorr
-          make mock
-          sleep 10
-          make test
+          make deploy
+# This should be re-enabled when Pocket IC supports threshold Schnorr
+# sleep 10
+# make test
--- a/motoko/threshold-schnorr/README.md
+++ b/motoko/threshold-schnorr/README.md
@@ -29,13 +29,16 @@ More specifically:
 This tutorial gives a complete overview of the development, starting with downloading [`dfx`](https://internetcomputer.org/docs/current/developer-docs/setup/index.md), up to the deployment and trying out the code on the mainnet.
 
 This walkthrough focuses on the version of the sample canister code written in
-Motoko programming language.. There is also a
+Motoko programming language. There is also a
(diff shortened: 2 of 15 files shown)


motoko/threshold-schnorr/README.md
```diff
@@ -57,14 +50,12 @@ This tutorial will use the Motoko version of the canister:
 cd examples/motoko/threshold-schnorr
 dfx start --background
 npm install
```
fspreiss, line 52: The README mentions `npm install` but the required `package.json` does not exist: calling `npm install` results in an error.

In the Github workflow this issue is circumvented by installing the required package manually with `npm install @noble/curves`.

We should either update the README, or add a `package.json` (and lock file). As far as I can see, many other examples in the repo do have a `package.json` (and also the respective `package-lock.json`).

.github/workflows/rust-threshold-schnorr-example.yml
```diff
@@ -25,11 +25,13 @@ jobs:
         run: bash .github/workflows/provision-pocket-ic-server.sh
       - name: Rust Threshold Schnorr Darwin
         run: |
+          dfxvm default 0.22.0-beta.0
```
fspreiss, line 28: Should we track removing this hardcoded version and re-enabling the disabled tests in a JIRA ticket?

fspreiss (approved):
Thanks, @altkdf!

.github/workflows/rust-threshold-schnorr-example.yml
andreacerulli: I don't know if I am doing anything wrong, but I if I follow the read me and try to sign I get `IC0302: Management canister has no method 'sign_with_schnorr`, which seems to indicate I am using the wrong version of dfx

fspreiss: This indeed indicates that the dfx version is too old.

@altkdf, should we adapt the READMEs accordingly?

I guess the README should say that at least dfx version `0.22.0-beta.0` is required, while pointing to [this](https://github.com/dfinity/dfxvm) regarding installing a specific dfx version, or directly giving the respective command:
```
DFX_VERSION=0.22.0-beta.0 sh -ci "$(curl -fsSL https://internetcomputer.org/install.sh)"
```

motoko/threshold-schnorr/README.md
altkdf: I think installing the dependency directly is currently okay because 1) it's the only `npm` dependency and 2) it's a test dependency. I removed most `npm install`s from READMEs though and added different scripts for deployment and testing. PTAL.

.github/workflows/rust-threshold-schnorr-example.yml
altkdf: I added to both READMEs that at least `0.22.0-beta.0` is required and added a ticket in the cleanup (CRP-2545) to track tests.

altkdf: Re hard-coded version: I'm not sure if that's a problem TBH. I actually think it might be even better because that's what we do for rust e.g. as well and makes tests reproducible. What we are missing IMO is something like `dfx default '>=0.22.0-beta.0'` with a lock file, but nothing like that exists so far, right?

motoko/threshold-schnorr/README.md
fspreiss: 👍

.github/workflows/rust-threshold-schnorr-example.yml
fspreiss: > Re hard-coded version: I'm not sure if that's a problem TBH. I actually think it might be even better because that's what we do for rust e.g. as well and makes tests reproducible.

There is an [Examples Maintenance Process](https://docs.google.com/document/d/1nQA3ISsZBo9HnwsNrZO4GqpSL1nCDxUu1PjQ_Yt0Gyo/edit) that says the following.

> Any time a new version of DFX is released (beta versions included), a new PR will be raised against the examples repo that updates the DFX_VERSION variable. Specifically, the files changed will be [provision-darwin.sh](https://github.com/dfinity/examples/blob/master/.github/workflows/provision-darwin.sh) and [provision-linux.sh](https://github.com/dfinity/examples/blob/master/.github/workflows/provision-linux.sh).

If we were to hard-code a DFX-version, this maintenance process would no longer catch issues. So, removing the hardcoded version should be tracked in a ticket.

altkdf: Ah, makes sense, didn't know that! Created https://dfinity.atlassian.net/browse/CRP-2546.

fspreiss: 👍

PR #946, opened 2024-07-26
Improve a comment in basic_bitcoin script spend

altkdf:


andreacerulli (approved):
Thanks for adding the clarification, LGTM!

PR #966, opened 2024-09-03
fix(rust): Update ic-cdk

bitdivine:
**Overview**
`ic-cdk` versions `0.13` and `0.14` have been yanked.

**Requirements**
Examples should not encourage users to use vulnerable versions of `ic-cdk`.

**Considered Solutions**
- Update `ic-cdk` version in all rust repositories.

**Recommended Solution**
- Update `ic-cdk` version in all rust repositories.

**Considerations**
There will be fewer crates that use this as an example that are then vulnerable. :-)

By the way, I note that old `ic-cdk` is also used by `pocket-ic` v3, used here, but `pocket-ic` is used only in development so this is less of a concern.  Updating `pocket-ic` is much harder than just bumping a version number.

### DIFF
--- a/.github/workflows/provision-pocket-ic-server.sh
+++ b/.github/workflows/provision-pocket-ic-server.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-POCKET_IC_SERVER_VERSION=${POCKET_IC_SERVER_VERSION:=4.0.0}
+POCKET_IC_SERVER_VERSION=${POCKET_IC_SERVER_VERSION:=5.0.0}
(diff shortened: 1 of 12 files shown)


mraszyk:
This [PR](https://github.com/dfinity/examples/pull/967) is enough to bump PocketIC. The example `rust/threshold-schnorr/src/schnorr_example_rust` won't work since it requires an unreleased version of PocketIC (and thus it doesn't work before this PR either).

gregorydemay (approved):
Thanks for the update!

PR #967, opened 2024-09-03
chore: bump PocketIC

mraszyk:


bitdivine:
Thanks!  Updating pocketic for the signer was rather harder!

PR #977, opened 2024-09-16
bump agent-js version

chenyan-dfinity:
* Bump @dfinity/agent to 2.1.3
* Update `send_http_get`, as the coinbase pro API is deprecated.


.github/workflows/update-package-json.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+# Find all package.json files in subdirectories
+find . -name "package.json" | while read -r file; do
+    # Check if the file contains any "@dfinity/" package
+    if grep -q '"@dfinity/' "$file"; then
+        # Use sed to replace the version for all @dfinity packages, preserving the caret if present
+        $SED_CMD -i 's/\("@dfinity\/[^"]*"\s*:\s*"\^*\)[0-9.]*"/\1'"$NEW_VERSION"'"/' "$file"
+        echo "Updated @dfinity/* versions in $file"
+    fi
```
rvanasa, line 34: If we want to update the lockfiles and include some error handling in the script:

```suggestion
    fi
    # Update `package-lock.json`
    npm install --prefix "$file/.." || exit 1
```

.github/workflows/update-package-json.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+    # Linux and other Unix-like systems
+    SED_CMD="sed"
+fi
+
+# Find all package.json files in subdirectories
+find . -name "package.json" | while read -r file; do
+    # Check if the file contains any "@dfinity/" package
+    if grep -q '"@dfinity/' "$file"; then
```
dfx-json, line 30: the only problem I see is that we assume all `@dfinity/` packages to be pinned to the same version. defining a list of known packages in this script and matching against that list might be a better approach

rvanasa: From the [agent-js repository](https://github.com/dfinity/agent-js/blob/main/package.json), this is the current list of relevant packages:
* `@dfinity/principal`
* `@dfinity/candid`
* `@dfinity/agent`
* `@dfinity/identity`
* `@dfinity/auth-client`
* `@dfinity/assets`
* `@dfinity/identity-secp256k1`
* `@dfinity/use-auth-client`

chenyan-dfinity: Right. I think the current assumption from agent-js is that `@dfinity/` all share the same version. Having a list of known packages would be more generic.

DSharifi (approved):

PR #1007, opened 2024-09-27
Fix Internet Identity integration in vetKD examples and make them work in Safari

fspreiss:
* Fixes the Internet Identity (II) integration in the vetKD examples by using the new environment variables for canister IDs (see [changelog of SDK 0.18](https://github.com/dfinity/sdk/blob/master/CHANGELOG.md#0180))
* Makes the example work in Safari by removing the `upgrade-insecure-requests;` content security policy (CSP) and by using `localhost` instead of `127.0.0.1` when calling II.

### DIFF
--- a/motoko/vetkd/src/app_frontend_js/src/index.js
+++ b/motoko/vetkd/src/app_frontend_js/src/index.js
@@ -235,10 +235,22 @@ async function ibe_decrypt(ibe_ciphertext_hex) {
 
 document.getElementById("login").onclick = async (e) => {
   e.preventDefault();
+
+  // According to https://github.com/dfinity/internet-identity?tab=readme-ov-file#local-replica,
+  // for local deployments, the II URL must be different depending on the browser:
+  // Chrome, Firefox: http://<canister_id>.localhost:4943
+  // Safari: http://localhost:4943?canisterId=<canister_id>
+  // 
+  // Safari detection rules are according to: https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#browser_name_and_version
+  let isSafari = /^(?!.*chrome\/\d+)(?!.*chromium\/\d+).*safari\/\d+/i.test(navigator.userAgent);
+  let identityProvider = isSafari ?
+    `http://localhost:4943/?canisterId=${process.env.CANISTER_ID_INTERNET_IDENTITY}` :
+    `http://${process.env.CANISTER_ID_INTERNET_IDENTITY}.localhost:4943/`;
+
   let authClient = await AuthClient.create();
   await new Promise((resolve) => {
     authClient.login({
-      identityProvider: `http://127.0.0.1:4943/?canisterId=${process.env.INTERNET_IDENTITY_CANISTER_ID}`,
+      identityProvider: identityProvider,
       onSuccess: resolve,
     });
   });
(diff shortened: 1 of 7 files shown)


motoko/vetkd/src/app_frontend_js/src/index.js
```diff
@@ -235,10 +235,22 @@ async function ibe_decrypt(ibe_ciphertext_hex) {
...
 document.getElementById("login").onclick = async (e) => {
   e.preventDefault();
+
+  // According to https://github.com/dfinity/internet-identity?tab=readme-ov-file#local-replica,
+  // for local deployments, the II URL must be different depending on the browser:
+  // Chrome, Firefox: http://<canister_id>.localhost:4943
+  // Safari: http://localhost:4943?canisterId=<canister_id>
+  // 
+  // Safari detection rules are according to: https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#browser_name_and_version
+  let isSafari = /^(?!.*chrome\/\d+)(?!.*chromium\/\d+).*safari\/\d+/i.test(navigator.userAgent);
```
frederikrothenberger, line 245: Where is that regex from? Just FYI: it does not match my user-agent on Safari, because that contains the `Safari` (capitalized).

fspreiss: I crafted it based on the rules for Safari recommended [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#browser_name_and_version) (See the comment one line above).

The capitalization of `Safari` shouldn't be a problem, because of the `i`nsensitive flag at the very end. Maybe you didn't add a version number, e.g., `/123` in your test?

Do you know of another/better regex that could be used?

rust/vetkd/src/app_frontend_js/src/index.js
```diff
@@ -235,10 +235,22 @@ async function ibe_decrypt(ibe_ciphertext_hex) {
...
+  let isSafari = /^(?!.*chrome\/\d+)(?!.*chromium\/\d+).*safari\/\d+/i.test(navigator.userAgent);
+  let identityProvider = isSafari ?
+    `http://127.0.0.1:4943/?canisterId=${process.env.CANISTER_ID_INTERNET_IDENTITY}` :
+    `http://${process.env.CANISTER_ID_INTERNET_IDENTITY}.localhost:4943/`;
+
   let authClient = await AuthClient.create();
   await new Promise((resolve) => {
     authClient.login({
-      identityProvider: `http://127.0.0.1:4943/?canisterId=${process.env.INTERNET_IDENTITY_CANISTER_ID}`,
+      identityProvider: identityProvider,
```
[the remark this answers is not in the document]
fspreiss: The respective team didn't update the reference example accordingly yet 😛

motoko/vetkd/src/app_frontend_js/src/index.js
frederikrothenberger: No the problem is the capitalization of `Safari`: I.e. your regex containing `safari` does not match `Safari`.

But in general I would use [ua-parser](https://www.npmjs.com/package/ua-parser-js).

rust/vetkd/src/app_frontend_js/src/index.js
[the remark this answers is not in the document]
frederikrothenberger: Fair enough 😉

frederikrothenberger (approved):
Thanks!

motoko/vetkd/src/app_frontend_js/src/index.js
fspreiss: Discussed offline. We'll keep the simple regex for now. Note for readers: ua-parser is [using](https://github.com/faisalman/ua-parser-js/blob/master/LICENSE.md) the AGPL license now (previously MIT).

PR #1010, opened 2024-09-30
fix: frontend button unresponsive

### LINKED ISSUE #992: Motoko minimal-counter-dapp: Frontend button does not work
Frontend button is unresponsive to input

marc0olo:
fixes #992 

**Note**:
- I switched the project structure to match what is currently generated when using `dfx new`
- I added an `decrement` function and the example now provides more buttons in the frontend to interact with the backend canister
- I kept the generated `declarations` folder in the project. personally I'd tend to remove it and put it into the `.gitignore`
- I wanted to update the agent dependency to version `2.1.2` but this seems to be incompatible with the latest `dfx` version `0.23.0`


marc0olo:
not sure why the pipeline fails 🤔 

I changed `test` to be dependent on `upgrade`. probably I need to switch back to relying on `install`. I made that change because running `test` twice resulted in an error locally.

domwoe (approved):

marc0olo:
@domwoe shall I aim to fix the darwin job or just merge the PR? it isn't the same issue, but here we also have problems with the darwin job in other PRs. most of them got merged and the issue was ignored. darwin provisioning files in other jobs, but somehow this is not an issue for this example. still it failes for whatever reason when trying to build the frontend.

wondering if there is a way to avoid the provisioning step in all of the jobs since it wastes a lot of resources 🤔

marc0olo:
I will just merge it now so that users have a working example again. let's address the github actions separately if we want that.

PR #1011, opened 2024-10-02
fix[995]: Delete motoko/http_counter directory

rumenov:
The example is pretty outdated. It is in the bottom 40 samples and I don't think it provides any value compared in addition to what we already have.


mraszyk:
Agreed! In particular, because this example does not certify the HTTP requests and thus does not follow the HTTP gateway protocol.

mraszyk (approved):

dfx-json (approved):

PR #1015, opened 2024-10-07
reenable tests and adjust cycles in rust threshold Schnorr

altkdf:
Pocket IC now supports threshold BIP340 and Ed25519, so the tests can be (re)enabled.

Pocket IC has two (implementation) limitations though:
1) It does not support BIP340 messages that are not 32 bytes long.
2) There is a bug that renames the `dfx_test_key`, so we have to temporarily use another key name, which is not very important though.

This PR also adjusts cycles attached to the `sign_with_schnorr` call. The adjusted amount of cycles is sufficient to make a call to a 34-node subnet, which is currently the case on mainnet for the production key. Note that overpaid cycles are reimbursed to the caller, which would be relevant for calls to smaller subnets like currently the case for the production test key.

Closes CRP-2585.

Closes CRP-2604.


fspreiss (approved):
Thanks, @altkdf!

PR #1024, opened 2024-10-16
feat(basic_bitcoin): cache ecdsa

dantol29:
**Overview**

The current process for retrieving a public key (via schnorr_public_key or ecdsa_public_key) takes approximately 9 seconds each time a user requests an address.  By storing public keys in a static variable, I want to improve performance and minimize cycle burn rate in the canister. It can serve as a good example for developers.

**Requirements**

Public keys should be stored in a static variable, allowing retrieval without calling the key generation functions every time.

**Considered Solutions**

thread_local storage with static variables that do persist across calls, but not across updates. 

**Recommended Solution**

Storing public keys in a static variable with HashMap for different derivation_paths

**Considerations**

This solution will significantly improve performance and lower cycle burn rate


altkdf (approved):
Thanks, LGTM!

altkdf:
@dantol29 I believe that the failing darwin job is something unrelated and has been fixed recently https://github.com/dfinity/examples/pull/1025. So the CI should become green if you merge `master`.

altkdf:
github seems confused about the last commits, maybe `git checkout cache-ecdsa && git merge master` would help?

PR #1028, opened 2024-10-22
feat(vetkd): CRP-2602 implement password manager using VetKD

altkdf:


altkdf:
Most of the files are currently copy-pasted from the encrypted notes example. The relevant changes are currently under `src/password_manager_rust`.

PR #1035, opened 2024-11-04
fix: await fetchRootKey in agent-js

mraszyk:
This PR fixes calls to `fetchRootKey` in agent-js to be awaited before the agent is actually used. Otherwise, signatures might be verified with the ICP mainnet root key before the root key is successfully fetched.


mraszyk:
https://dfinity.slack.com/archives/CGA566TPV/p1730739810179459

mraszyk:
The fix should be implemented in agent-js.

PR #1042, opened 2024-11-11
feat(crypto): CRP-2624 Move tECDSA/tSchnorr/vetKD fake canister to separate repo

fspreiss:
We decided to move the tECDSA/tSchnorr/vetKD mock/fake canister to a separate repository, as it is not really an example dapp, so this example dapp is deleted again. Originally, the code was added in https://github.com/dfinity/examples/pull/1029.

The new repository is https://github.com/dfinity/chainkey-testing-canister.

PR #1055, opened 2024-11-22
docs: fix links

marc0olo:
- wrong LICENSE file name/reference lead to a broken link in the portal URL check
- github issue link didn't make sense


altkdf:
Thanks for looking into this @marc0olo. This example will be (re)moved though https://github.com/dfinity/examples/pull/1042.

marc0olo:
@altkdf thanks for pointing that out. mhh, can we somehow get the conflicts in the other branch fixed and get it merged as soon as possible?

otherwise it would be great to just approve this now so that we are unblocked in terms of getting the docs updated in the portal.

I do not think it is very urgent, so I am also fine just closing this PR here 😅 

I didn't know about the sideeffects with the portal repo. so we should always be careful when linking between markdowns here in the examples repo

altkdf (approved):
Then let's just approve this PR :)

PR #1062, opened 2024-11-28
chore: Add Pocket-IC tests to the parallel calls example CI

oggy-dfin:
The parallel calls Rust example demonstrate the performance difference between sequential and parallel calls using Pocket IC, but this hasn't been tested in CI so far (and in fact the example stopped working as expected after the cycles costs for inter-canister calls were increased).

### DIFF
--- a/.github/workflows/rust-parallel-calls-example.yml
+++ b/.github/workflows/rust-parallel-calls-example.yml
@@ -42,3 +42,26 @@ jobs:
           dfx start --background
           make test
           popd
+  rust-parallel-calls-pocket-ic-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Provision Linux
+        run: bash .github/workflows/provision-linux.sh
+      - name: Install PocketIC server
+        uses: dfinity/pocketic@main
+      - name: Test parallel calls with PocketIC
+        run: |
+          pushd rust/parallel_calls
+          cargo build --release --target wasm32-unknown-unknown -p callee
+          cargo build --release --target wasm32-unknown-unknown -p caller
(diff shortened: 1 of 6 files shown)


.github/workflows/rust-parallel-calls-example.yml
```diff
@@ -42,3 +42,26 @@ jobs:
...
+      - name: Test parallel calls with PocketIC
+        run: |
+          pushd rust/parallel_calls
+          cargo build --release --target wasm32-unknown-unknown -p callee
+          cargo build --release --target wasm32-unknown-unknown -p caller
+          export CALLER_WASM=$(cargo build --target wasm32-unknown-unknown -p caller --message-format=json \
+            | jq -r 'select(.reason == "compiler-artifact") | .filenames[] | select(endswith(".wasm"))')
+          export CALLEE_WASM=$(cargo build --target wasm32-unknown-unknown -p callee --message-format=json \
+            | jq -r 'select(.reason == "compiler-artifact") | .filenames[] | select(endswith(".wasm"))')
+          cargo run
```
mraszyk, line 65: using cargo test (and `#[test]` instead of a main function) is more common

mraszyk (approved):

.github/workflows/rust-parallel-calls-example.yml
oggy-dfin: Agreed in principle but I wrote this test to be run from the command line (as specified in the `README`). Doing this as a test would make it a bit more complicated for the user (they'd have to pass flags to see the output, put the `pocket-ic` binary in the right directory etc). So I think it's a bit easier for them to keep it as a binary.

mraszyk: > put the pocket-ic binary in the right directory etc

this is necessary anyway, but I get your point with "--nocapture"

PR #1068, opened 2024-12-11
fix: drop Host header in https outcall request

ninegua:
Having the Host header seems to cause problem with https outcalls using http/2, which is currently the case.

It is best to remove this setting from examples because many people (me included) would copy them and later have a hard time figuring out what could be the problem if something goes wrong.


marc0olo (dismissed):
agree. thanks @ninegua!

marc0olo:
@ninegua seems like you need to resolve some conflicts first 😅

marc0olo (approved):

PR #1069, opened 2024-12-13
Examples for the new Rust tutorial on inter-canister calls

oggy-dfin:
DO NOT MERGE until best-effort responses are available on mainnet.

This is a draft that I can't currently test, as it seems that `msg_deadline` has gone AWOL in the process of overhauling Rust CDK's next branch, but putting it up to get feedback before Christmas.

**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


rust/best-effort-response-calls/src/berc_backend/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    ic_cdk::api::call::msg_deadline()
+}
+
+#[ic_cdk::query(composite = true)]
+async fn deadline_in_composite_query() -> u64 {
+    ic_cdk::api::call::msg_deadline()
+}
+
+#[ic_cdk::query(composite = true)]
+async fn test_deadlines_in_composite_query() -> (u64, u64) {
```
alin-at-dfinity, line 80: Do we actually do anything at all with deadlines in composite queries, beyond exposing them to the callee? Should we?

oggy-dfin: We ended up [specifying this](https://github.com/dfinity/interface-spec/pull/268/files#diff-b9a833d6114611a080aee4cd57618feb45aa044bcc11105d44b155b5bbde00ebR1563):
> For other calls (ingress messages and all calls to query and composite query methods, including calls in replicated mode) a deadline of 0 is returned.

The implementation is consistent with that. I've added this to the README now.

jessiemongeon1 (approved):

jennifertrin (approved):

PR #1071, opened 2024-12-19
add BIP341 to the threshold Schnorr rust example

altkdf:


fspreiss (dismissed):

rust/threshold-schnorr/src/schnorr_example_rust/schnorr_example.did
```diff
@@ -1,7 +1,8 @@
 type SchnorrAlgorithm = variant {bip340secp256k1; ed25519};
 
 service : {
+  for_test_only_change_management_canister_id : (text) -> (variant { Ok: null; Err: text });
```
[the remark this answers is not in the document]
altkdf: Yes, created [CRP-2664](https://dfinity.atlassian.net/browse/CRP-2664). Adding TODOs IMO is not really necessary here because easy to find.

rust/threshold-schnorr/src/schnorr_example_rust/tests/tests.rs
```diff
@@ -37,35 +62,74 @@ fn test_impl(pic: &PocketIc, algorithm: SchnorrAlgorithm) {
...
+        "for_test_only_change_management_canister_id",
+        encode_one(schnorr_mock_canister_id.to_text()).unwrap(),
+    )
+    .expect("failed to update management canister id");
+    // Make sure the example canister uses mock schnorr canister instead of
+    // the management canister
+    fast_forward(&pic, 5);
+
     // a message we can reverse to break the signature
     // currently pocket IC only supports 32B messages for BIP340
```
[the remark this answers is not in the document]
altkdf: Good point. That doesn't seem to be the case, so I removed the comment. But 32B still seems a good choice for the message length. And since the tests already take a few minutes, we probably shouldn't add more different message lengths, especially because we are already testing this API elsewhere.

rust/threshold-schnorr/src/schnorr_example_rust/tests/tests.rs
```diff
@@ -37,35 +62,74 @@ fn test_impl(pic: &PocketIc, algorithm: SchnorrAlgorithm) {
...
-    let public_key_hex = pk_reply.unwrap().public_key_hex;
+    if sig_reply.is_err() {
+        assert!(!should_validate);
+        let dummy_signature_hex = String::from("a".repeat(64));
+        assert_ne!(
+            update(
+                &pic,
+                my_principal,
+                example_canister_id,
+                "verify",
```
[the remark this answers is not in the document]
altkdf: Yeah, that's not very obvious, added the following.
```rust
        // If we failed to produce a signature with particular testing
        // parameters, still test that the verification fails on dummy inputs.
```

rust/threshold-schnorr/src/schnorr_example_rust/tests/tests.rs
```diff
@@ -110,12 +180,13 @@ fn test_impl(pic: &PocketIc, algorithm: SchnorrAlgorithm) {
...
                 clone_and_reverse_chars(&message),
                 public_key_hex.clone(),
+                merkle_tree_root_hex.clone(),
                 algorithm,
             ))
             .unwrap(),
         );
 
-        assert!(!verification_reply.unwrap().is_signature_valid);
+        assert_ne!(verification_reply, successful_validation.clone());
```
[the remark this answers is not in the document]
altkdf: Here yes, changed. In 230-233 we expect an error but no a validation returning `false`. So this would need a `match`. And since this would be the only place where `assert_matches` would be used, I would prefer to leave it as is.

rust/threshold-schnorr/src/schnorr_example_rust/build.sh
```diff
@@ -0,0 +1,21 @@
...
+set -euo pipefail
+
+TARGET="wasm32-unknown-unknown"
+CANISTER="schnorr_example_rust"
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+pushd $SCRIPT_DIR
+
+# NOTE: On macOS a specific version of llvm-ar and clang need to be set here.
+# Otherwise the wasm compilation of rust-secp256k1 will fail.
```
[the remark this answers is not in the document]
randombit: Re `secp256k1` vs `k256` yes `bitcoin` depends directly on `secp256k1`, which is a binding to the C library. It seems like (for reasons I do not currently understand) the way `secp256k1` builds you can either be compatible with wasm *or* compatible with normal native applications, but not both.

rust/threshold-schnorr/src/schnorr_example_rust/src/wasm_only.rs
```diff
@@ -118,6 +186,43 @@ fn verify_bip340_secp256k1(
...
+            .0
+            .to_inner()
+            .serialize()
+    };
+
+    let sig =
+        k256::schnorr::Signature::try_from(sig_bytes).expect("failed to deserialize signature");
+
+    let vk = k256::schnorr::VerifyingKey::from_bytes(&tweaked_pk_bytes)
+        .expect("failed to deserialize BIP340 encoding into public key");
```
randombit, line 219: IIRC we can do the entire verification using `secp256k1` here rather than having to use both `k256` and `secp256k`.

randombit: Alternately, with a bit more effort (copying the code from https://github.com/dfinity/ic/blob/master/rs/crypto/secp256k1/src/lib.rs#L352) we could do the verification using only `k256` and avoid the `bitcoin` dependency. Given the apparent complications in the build that this brings in, it might be worthwhile.

fspreiss (approved):
Re-approving because review was automatically dismissed

rust/threshold-schnorr/src/schnorr_example_rust/src/wasm_only.rs
altkdf: Removed `k256`.

Re removing `bitcoin`, there's also another aspect, which is complexity. I guess here, among other things, we want to show how simple it is to use threshold Schnorr. It's useful to show that it's perfectly fine to use the existing `bitcoin` crate with threshold Schnorr in a canister.

rust/threshold-schnorr/README.md
```diff
@@ -191,21 +211,58 @@ For obtaining the canister's root public key, the derivation path in the API can
 
 Computing threshold Schnorr signatures is the core functionality of this feature. **Canisters do not hold Schnorr keys themselves**, but keys are derived from a master key held by dedicated subnets. A canister can request the computation of a signature through the management canister API. The request is then routed to a subnet holding the specified key and the subnet computes the requested signature using threshold cryptography. Thereby, it derives the canister root key or a key obtained through further derivation, as part of the signature protocol, from a shared secret and the requesting canister's principal identifier. Thus, a canister can only request signatures to be created for its canister root key or a key derived from it. This means, that canisters "control" their private Schnorr keys in that they decide when signatures are to be created with them, but don't hold a private key themselves.
 
+The threschold Schnorr signature API allows to pass auxiliary information for
```
fspreiss, line 214: nit: threschold -> threshold

fspreiss (approved):
Re-approving because review was automatically dismissed

rust/threshold-schnorr/README.md
altkdf: done

fspreiss (approved):
Re-approving because review was automatically dismissed

rust/threshold-schnorr/README.md
```diff
@@ -83,27 +84,46 @@ To deploy this canister the mainnet, one needs to do two things:
 
 Deploying to the Internet Computer requires [cycles](https://internetcomputer.org/docs/current/developer-docs/getting-started/tokens-and-cycles) (the equivalent of "gas" on other blockchains).
 
+#### Update management canister ID reference for testing
+
+The latest version of `dfx`, `v0.24.3`, does not yet support
+`opt_merkle_tree_root_hex` that is not `None`. Therefore, for local tests, [the
+chainkey testing canister](https://github.com/dfinity/chainkey-testing-canister)
```
jessiemongeon1, line 91:
```suggestion
chain-key testing canister](https://github.com/dfinity/chainkey-testing-canister)
```

rust/threshold-schnorr/README.md
```diff
@@ -83,27 +84,46 @@ To deploy this canister the mainnet, one needs to do two things:
...
+can be installed and used instead of the management canister. Note also that the
+chainkey testing canister is deployed on mainnet and can be used for mainnet
```
jessiemongeon1, line 93:
```suggestion
chain-key testing canister is deployed on the mainnet and can be used for mainnet
```

rust/threshold-schnorr/README.md
```diff
@@ -83,27 +84,46 @@ To deploy this canister the mainnet, one needs to do two things:
...
+testing to reduce the costs, see the linked repo for more details.
+
+This sample canister allows the caller to change the management canister address
+for Schnorr by calling the `for_test_only_change_management_canister_id`
+endpoint with the target canister principal. With `dfx`, this can be done
+automatically with `make mock`, which will install the chainkey testing canister
```
jessiemongeon1, line 99:
```suggestion
automatically with `make mock`, which will install the chain-key testing canister
```

jessiemongeon1 (approved):

altkdf:
Oh, approvals are reset even if I only apply suggestions from the PR via the github UI. That's quite extreme...

jessiemongeon1 (approved):

fspreiss (approved):
Re-approving because review was automatically dismissed

PR #1079, opened 2025-01-09
update codeowner of iii example

marc0olo:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -21,7 +21,7 @@
 /motoko/hello_cycles/ @dfinity/languages
 /motoko/ic-pos/ @dfinity/growth
 /motoko/icrc2-swap/ @dfinity/growth
-/motoko/internet_identity_integration/ @dfinity/gix
+/motoko/internet_identity_integration/ @dfinity/identity
 /motoko/life/ @dfinity/languages
 /motoko/minimal-counter-dapp/ @dfinity/growth
 /motoko/parallel_calls/ @dfinity/languages


sea-snake (dismissed):

.github/CODEOWNERS
```diff
@@ -21,7 +21,7 @@
 /motoko/hello_cycles/ @dfinity/languages
 /motoko/ic-pos/ @dfinity/growth
 /motoko/icrc2-swap/ @dfinity/growth
-/motoko/internet_identity_integration/ @dfinity/gix
+/motoko/internet_identity_integration/ @dfinity/identity
```
cgundy, line 24: try
```suggestion
/motoko/internet_identity_integration/ @dfinity/Identity
```

marc0olo: I didn't expect this to be case sensitive 😅  will test

marc0olo: @cgundy seems like the issue still applies. does the team have write access?

cgundy: No they don't, that's probably it! Let me give them write access and see if that changes things

cgundy: And looks like it's lowercase after all

marc0olo: ok now it is valid. wondering if that was a caching issue on Github 🤔

marc0olo: thanks, will revert to lowercase then 👍

sesi200 (approved):

PR #1074, opened 2024-12-20
add BIP341 to the `basic_bitcoin` rust example

altkdf:
Also restructures the types of P2TR addresses. The reason for the restructuring is that the addition of BIP341 allows for a more flexible usage of P2TR.

Before:
1) Untweaked key path address and spend.
2) Script path address and spend.

Now:
1) Tweaked key path address with unspendable merkle root (as recommended by BIP341).
2) P2TR address with a key path spend and a script path spend, where we have a different key for each spending path (both using threshold Schnorr though).

Tested locally with regtest.

PR #1080, opened 2025-01-10
add BIP341 to Motoko `basic_bitcoin`

altkdf:


motoko/basic_bitcoin/src/basic_bitcoin/src/EcdsaApi.mo
```diff
@@ -8,45 +8,38 @@ module {
...
+  type EcdsaCanisterActor = Types.EcdsaCanisterActor;
 
   // The fee for the `sign_with_ecdsa` endpoint using the test key.
   let SIGN_WITH_ECDSA_COST_CYCLES : Cycles = 10_000_000_000;
 
-  let ecdsa_canister_actor : EcdsaCanisterActor = actor("aaaaa-aa");
-
   /// Returns the ECDSA public key of this canister at the given derivation path.
-  public func ecdsa_public_key(key_name : Text, derivation_path : [Blob]) : async Blob {
+  public func ecdsa_public_key(ecdsa_canister_actor: EcdsaCanisterActor, key_name : Text, derivation_path : [Blob]) : async Blob {
```
altkdf, line 17: Since `dfx`/ pocket IC doesn't support key tweaking yet, we need to use the mock canister. Previously, the actor was defined as `actor ("aaaaa-aa")` in the module, but now we need to change it dynamically based on the canister API, but Motoko doesn't allow for non-static variables in modules. Therefore, passing `EcdsaCanisterActor` as an argument is the best way to make it work that I could find.

altkdf:
Motoko style actually uses camel case for variables and function names, but I will leave this change for another PR...

motoko/basic_bitcoin/src/basic_bitcoin/src/P2tr.mo
```diff
@@ -0,0 +1,427 @@
...
+                };
+            };
+            // Verify that our own address is P2TR key spend address.
+            case (#err msg) Debug.trap("This example supports signing p2tr key spend addresses only: " # msg);
+        };
+
+        transaction.toBytes();
+    };
+
+    /// Sign a bitcoin transaction.
```
fspreiss, line 187: nit: Both `sign_key_spend_transaction` and `sign_script_spend_transaction` have the documentation "Sign a bitcoin transaction." which seemingly comes from the old code. Also, one func is public, while the other is not. Could both be non-public?

fspreiss (approved):
Thanks, @altkdf! LGTM, [not an expert](https://github.com/dfinity/examples/pull/1074#pullrequestreview-2532242721) on the topic, however. Quite some code was moved to new files: I only glanced over this code and didn't do an in-depth review.

motoko/basic_bitcoin/src/basic_bitcoin/src/EcdsaApi.mo
altkdf: Not necessarily. It seems useful for testing to be able to change the management canister actor easily.

rust/threshold-schnorr/src/schnorr_example_rust/src/wasm_only.rs
```diff
@@ -146,17 +146,17 @@ async fn verify(
     algorithm: SchnorrAlgorithm,
 ) -> Result<SignatureVerificationReply, String> {
     let sig_bytes = hex::decode(&signature_hex).expect("failed to hex-decode signature");
-    let msg_bytes = message.as_bytes();
+    let msg_bytes = hex::decode(&message).expect("failed to hex-decode message");
```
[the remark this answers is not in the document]
altkdf: Ah, I used that canister for debugging to check that the signatures are valid because the implementation didn't work out of the box. I forgot to revert the changes. Reverted now.

motoko/basic_bitcoin/src/basic_bitcoin/src/P2tr.mo
altkdf: I added a few words but that's not far away from the function names themselves... The reason why  `sign_key_spend_transaction` is public is because it is reused in `P2trKeyOnly.mo`.

altkdf:
It would also make sense if someone except me would check that this example works with `bitcoind` because there are no automated tests.

berestovskyy (approved):
Looks good, thanks!

fspreiss (approved):

PR #1073, opened 2024-12-19
Update Rust project READMEs

jessiemongeon1:
The example README files are no longer used on the Dev Docs, so the keyword tags and format can be adjusted to read best on GitHub.

Each README has also been updated to fit latest style guide practices.

PR #1081, opened 2025-01-16
fix link to download latest icrc1-ledger

marc0olo:
I recognized we introduced a link which doesn't work in this PR that got merged recently: https://github.com/dfinity/examples/pull/1073

### DIFF
--- a/motoko/token_transfer/README.md
+++ b/motoko/token_transfer/README.md
@@ -40,7 +40,7 @@ The URL for the ledger .did file is `https://raw.githubusercontent.com/dfinity/i
 If you want to make sure, you have the latest ICRC-1 ledger files you can run the following script.
 
 ```sh
-curl -o download_latest_icrc1_ledger.sh "https://raw.githubusercontent.com/dfinity/ic/<REVISION>/rs/rosetta-api/scripts/download_latest_icrc1_ledger.sh"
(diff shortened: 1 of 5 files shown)


letmejustputthishere:
i think its better to point to a specific commit in case the file get's renamend or put into a different directory

marc0olo:
I think this is in general not very convenient at this point, but I see your point. IMO it is fine for the moment and we should rather consider to create a dedicated repository just for scripts to be (re-)used across different repos.

personally I prefer relying on master over some potentially very outdated commit. if the script needs to change for whatever reason, we ultimately have a similar issue.

marc0olo:
@letmejustputthishere I switched to a specific commit now. my concerns above remain though 😛

letmejustputthishere:
the issue is also present here
https://github.com/dfinity/examples/tree/master/motoko/token_transfer_from
https://github.com/dfinity/examples/tree/master/rust/icp_transfer
https://github.com/dfinity/examples/tree/master/rust/token_transfer
https://github.com/dfinity/examples/tree/master/rust/token_transfer_from

do you mind fixing them as well?

marc0olo:
@letmejustputthishere I think I only missed `https://github.com/dfinity/examples/tree/master/rust/icp_transfer` which I fixed now, too

letmejustputthishere:
thank you!

letmejustputthishere (approved):

PR #1086, opened 2025-01-24
fix(ICRC_Ledger): FI-1654: Update link to script that downloads latest ICRC ledger

mbjorkqvist:
The script that downloads the latest ICRC ledger `wasm` and `did` files was recently updated to download them using [the official ledger releases](https://github.com/dfinity/ic/releases?q=ledger-suite-icrc&expanded=false), rather than the latest commit on the `master` branch of the [ic](https://github.com/dfinity/ic) repository.

The links to the script in the documentation point to a specific commit in the `ic` repository. This PR proposes to update the links to the commit where the script was updated, so that users end up with an officially released ledger.

### DIFF
--- a/motoko/token_transfer/README.md
+++ b/motoko/token_transfer/README.md
@@ -40,7 +40,7 @@ The URL for the ledger .did file is `https://raw.githubusercontent.com/dfinity/i
 If you want to make sure, you have the latest ICRC-1 ledger files you can run the following script.
 
 ```sh
-curl -o download_latest_icrc1_ledger.sh "https://raw.githubusercontent.com/dfinity/ic/1f1d8dd8c294d19a5551a022e3f00f25da7dc944/rs/rosetta-api/scripts/download_latest_icrc1_ledger.sh"
+curl -o download_latest_icrc1_ledger.sh "https://raw.githubusercontent.com/dfinity/ic/69988ae40e4cc0db7ef758da7dd5c0606075e926/rs/rosetta-api/scripts/download_latest_icrc1_ledger.sh"
 chmod +x download_latest_icrc1_ledger.sh
 ./download_latest_icrc1_ledger.sh
 ```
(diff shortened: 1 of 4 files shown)


marc0olo (approved):
I just want to refer to a recent PR and discussion in that regard where I fixed the link and general and wanted to switch to master for the download script 😅 

see https://github.com/dfinity/examples/pull/1081

I assume this would have prevented the need for this PR. not sure if there is a really good path to tackle this. maybe there should be a more common place for the script to be stored along with other useful scripts 🤔

marc0olo (approved):
@mbjorkqvist just to be sure, the ICP ledger download script did not change, right?

mbjorkqvist:
Thanks for the review and feedback, @marc0olo!

> I just want to refer to a recent PR and discussion in that regard where I fixed the link and general and wanted to switch to master for the download script 😅
> 
> see https://github.com/dfinity/examples/pull/1081
> 
> I assume this would have prevented the need for this PR. not sure if there is a really good path to tackle this. maybe there should be a more common place for the script to be stored along with other useful scripts 🤔

Thanks for the pointer! I'd rather point to a specific commit in this case, since there is a non-negligible chance that the files get renamed or moved, in which case the user may have a hard time finding the correct one. In fact, the script would IMO fit better under `rs/ledger_suite/scripts` than under `rs/rosetta-api/scripts`, so it may move in the near future, but that would be part of a somewhat larger cleanup item, so I don't want to move it now.

> @mbjorkqvist just to be sure, the ICP ledger download script did not change, right?

No, it didn't change (yet). We are planning to create ICP ledger suite releases similar to the ones we now have for [the ICRC ledger suite](https://github.com/dfinity/ic/releases?q=ledger-suite-icrc&expanded=false), starting from the next ICP ledger suite version, which is planned for the next week or two, in which case we'd want to update that script also. We have the [FI-1647](https://dfinity.atlassian.net/browse/FI-1647) JIRA ticket for updating the script when the release is done.

letmejustputthishere:
@mbjorkqvist can you please ping us so we can make sure to update the examples READMEs as well. @marc0olo did you update the examples with the current commit hash already, or should we wait for two weeks until FI provides the new scripts?

PR #1101, opened 2025-02-25
Add security disclaimer in the icrc-2 swap example

eduarddfinity:
This PR adds a security disclaimer to the ICRC-2 swap example to highlight potential security risks in inter-canister calls. It also includes a link to the security best practices documentation, where these issues are explained in more detail.


marc0olo (approved):
note: created `GROWTH-139` to improve the example and avoid the async bug trap

PR #1102, opened 2025-02-26
chore: Fixing the ECDSA test key name for local deployment of the basic Bitcoin dapp

THLO:
The key `dfx_test_key` should be used for local development of Bitcoin dapps using `regtest` mode. This was changed when working on threshold Schnorr. As Schnorr signatures are supported now, the key name can be reverted to `dfx_test_key`.


letmejustputthishere (approved):

altkdf (approved):
Thanks!

dsarlis (approved):

PR #1110, opened 2025-03-18
chore: migrate Ninja projects here

sesi200:


motoko/hello_world/backend/app.mo
```diff
@@ -1,4 +1,4 @@
-actor HelloWorld {
+persistent actor HelloWorld {
   // We store the greeting in a stable variable such that it gets persisted over canister upgrades.
   stable var greeting : Text = "Hello, ";
```
fxgst, line 3:
```suggestion
  var greeting : Text = "Hello, ";
```

jessiemongeon1 (approved):

fxgst (commented):
Can we use a similar workflow like in the ninja repo such that we dont need to create a new workflow file for each project?

.github/workflows/hosting-my-crypto-blog-example.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Provision Darwin
+        run: bash .github/workflows/provision-darwin.sh
+      - name: Hosting My Crypto Blog Darwin
+        run: |
+          dfx start --background
+          pushd hosting/my_crypto_blog
+          make install
+          popd
```
fxgst, line 28: why do we use make here? Also, I think we can drop building it on macOS, but instead build it inside the dev container on linux

fxgst (approved):
LGTM!

PR #1123, opened 2025-04-04
chore: Update the composite_query example to work with the new cdk

raymondk:
**Overview**
The example was using an older version of the cdk and did not work with the new cycles requirements for creating a canister.

**Requirements**
A user can follow the instructions in the README and get the expected results.


berestovskyy (approved):
Looks good, thanks!

jessiemongeon1 (approved):

PR #1124, opened 2025-04-04
chore: Fix the port number of the canister url

raymondk:
**Overview**
The README is inaccurate based on the dfx.json in this project.


jessiemongeon1 (approved):

jessiemongeon1:
The screenshot image will also need to be updated to reflect the correct port.

PR #1126, opened 2025-04-05
chore: use the standard port for the random_maze

raymondk:
**Overview**
The random_maze project did not work according to the instructions.

* Local dev server proxies to the default local replica port


crusso (approved):
Rubber stamping

PR #1127, opened 2025-04-05
chore: use standard port for the life example

raymondk:
**Overview**
The life example did not work according to the instructions.

* Local dev server proxy defaults to the local replica port


crusso (approved):
Rubber stamping

PR #1129, opened 2025-04-06
chore: use standard port for cert-var example

raymondk:
**Overview**
The cert-var example had the local dev server proxy to the old port.

* Bindings were not consistent with the code that dfx generate produces in newer versions
* Local dev server proxy defaults to the local replica port


crusso (approved):
rubber stamping

PR #1131, opened 2025-04-24
update: static website README

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/hosting/static-website/README.md
+++ b/hosting/static-website/README.md
@@ -1,24 +1,16 @@
 ---
+
 keywords: [static website, basic website, html, host a website, beginner]
+
 ---
 
 # Static website
 
 [View this sample's code on GitHub](https://github.com/dfinity/examples/tree/master/hosting/static-website)
 
-## Overview
-The example shows how to deploy a simple, static website on the IC in an asset canister. The website is very simple, it just shows the DFINITY logo and has a basic styling of the logo. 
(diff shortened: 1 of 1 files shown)


hosting/static-website/README.md
```diff
@@ -100,28 +80,51 @@ The `dfx.json` file is a configuration file which specifies the canister used fo
...
 
+```
 
-## Testing
 The URL for the frontend depends on the canister ID.  Local canister IDs can be obtained using `dfx canister id`, in this case `dfx canister id www`. When deployed, the URL will look like this:
 
-**http://\{ui_canister_id\}.localhost:4943**
+```
+
+http://\{canister_id\}.localhost:4943
```
ilbertt, line 107: Do we need those backslashes?

```suggestion
http://{canister_id}.localhost:4943
```

Feel free to resolve

ilbertt (approved):
Approving, with 2 observations + 1 comment in the code:

- it would be nicer to have the Internet Computer logo in the assets, instead of the DFINITY logo, just for branding purposes
- the screenshot shows the `http://{canister_id}.localhost:8000` URL instead of the `http://{canister_id}.localhost:4943` URL written in the README.

This would also be a good occasion to re-take the screenshot in order to remove the browser's _Update_ notification on the top right.

jessiemongeon1:
Thanks @ilbertt , the example will likely be revised in the near future, but this PR is to address some immediate feedback from Jan.

PR #1144, opened 2025-05-12
Migrate Rust vetKD example to use management canister

fspreiss:
Migrates the Rust vetKD example to directly use the management canister instead of the system API mock.


altkdf (approved):
Also LGTM! Essentially the same comment as in the Motoko PR. Maybe we don't need the 2 extra steps in general.

rust/vetkd/src/app_backend/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-candid = "0.8.4"
-ic-cdk = "0.7.1"
+candid = "0.10.13"
+ic-cdk = "0.18.1"
```
altkdf, line 13: FYI https://forum.dfinity.org/t/yanked-ic-cdk-v0-18-investigating-reported-issues/47908

fspreiss:
> Also LGTM! Essentially the same comment as in the Motoko PR. Maybe we don't need the 2 extra steps in general.

[Done](https://github.com/dfinity/examples/pull/1144/commits/de76beaac327936f5fab1e2a907234b35d01220c). See also https://github.com/dfinity/examples/pull/1145#discussion_r2088912816.

rust/vetkd/src/app_backend/Cargo.toml
altkdf: going with `0.18.2` sounds good to me :+1:

PR #1145, opened 2025-05-12
Migrate Motoko vetKD example to use management canister

fspreiss:
Migrates the Motoko vetKD example to directly use the management canister instead of the system API mock.

Also updates the README, cleans the package.json, updates Rust dependencies, and updates the Content Security Policy (CSP).


motoko/vetkd/README.md
```diff
@@ -1,51 +1,46 @@
...
-```bash
-dfx canister create system_api --specified-id s55qq-oqaaa-aaaaa-aaakq-cai
+```sh
+cd examples/motoko/vetkd
+npm install
 ```
 
-Without this, `dfx` may use different canister IDs for the `system_api` and `app_backend` canisters in your local environment.
-
-## Step 3: Register, build, and deploy the project:
```
altkdf, line 48: Are steps 3 and 5 supposed to be empty? Also, does one really need to open a new terminal window?

altkdf (commented):
Thanks @fspreiss , LGTM, just some minor comments in the README.

altkdf (approved):

motoko/vetkd/README.md
fspreiss: Ah, I copied the text from the Rust example, where only the `--clean` flag was used, but not the `--background` one. I adapted both READMEs now to use both `--background --clean` and removed the step to open a second terminal.

Some steps indeed were supposed to be empty, but that indeed looks weird, so I changed it so that they are not empty.

PR #1148, opened 2025-05-19
Chore/bump-ic-cdk-version

letmejustputthishere:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


rust/send_http_get/src/send_http_get_backend/Cargo.toml
```diff
@@ -9,10 +9,7 @@ edition = "2021"
...
 [dependencies]
-candid = "0.8.2"
-ic-cdk = "0.6.0"
-ic-cdk-macros = "0.6.0"
-serde = "1.0.152"
-serde_json = "1.0.93"
-serde_bytes = "0.11.9"
-
+candid = "0.10.14"
+ic-cdk = "0.18.2-alpha.1"
```
marc0olo, line 13:
```suggestion
ic-cdk = "0.18.2"
```

I would prefer waiting for this to be unyanked and waiting with merging this PR until that happens.

PR #1153, opened 2025-06-04
update llm_chatbot to v1 endpoint

davencyw:
**Overview**
This update moves away from the deprecated v0 endpoint and updates the LLM library which uses the newer v1 endpoint that also allows to call tools.

**Requirements**
No requirements

**Considered Solutions**
We could leave it unchanged, but that would prevent people from accessing the benefits of the newer v1 endpoint.

**Recommended Solution**
Updating the library and the function calls.

**Considerations**
No impact.


letmejustputthishere (approved):
workflow seems to fail in the CI, i can deploy locally though

fxgst:
I'm working on it in a separate PR

fxgst:
Hi @davencyw, you should be able to merge master into your branch and CI should pass! We found a [workaround](https://github.com/dfinity/examples/pull/1154).

PR #1156, opened 2025-06-06
chore: Remove basic DAO Rust example, move canister-info example to execution

fxgst:


mraszyk (approved):

mraszyk:
Why do we only remove the Rust example while keeping the Motoko counterpart?

letmejustputthishere:
instead of removing it, could we move it to `archive`?

ggreif (approved):
LGTM!

PR #1160, opened 2025-06-06
CI test 1, do not merge

fxgst:


.github/workflows/ninja_pr_checks.yml
```diff
@@ -101,8 +101,10 @@ jobs:
...
       - name: Start dfx and build project
-        working-directory: ${{ matrix.example.path }}
         run: |
           # Note: Running these commands in the same step and with verbose output magically fixes the 502 Bad Gateway issue in dfx/pocketic for the LLM Chatbot example
           RUST_LOG=trace dfx start --background --verbose
-          dfx deploy
+
+      - name: Deploy
+        working-directory: ${{ matrix.example.path }}
+        run: dfx deploy
```
mraszyk: Could you specify which canister should be deployed here (and omit dependencies in dfx.json)?

PR #1179, opened 2025-06-17

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/motoko/icp_transfer/README.md
+++ b/motoko/icp_transfer/README.md
@@ -1,9 +1,5 @@
 # ICP transfer
 
-[View this samples code on GitHub](https://github.com/dfinity/examples/tree/master/motoko/icp_transfer).
-
-## Overview
-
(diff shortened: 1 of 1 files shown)


marc0olo (commented):
@jessiemongeon1 I think this example is a little bit tricky. I am not sure how valuable this example is in conjunction with icp.ninja. IMO, there is a reason this readme included detailed steps to be executed locally.

can we maybe distinguish between examples which are designed to be used locally only and those which make sense to be used/explored in icp.ninja? (cc @letmejustputthishere)

letmejustputthishere:
the steps in the readme tell the reader how we got to the example. i think this info is nowadays also available in the developer docs, but we should at least point to it

marc0olo:
mhh, I don't think that I would like to explore this example in ICP.ninja because it requires real ICP due to the use of the remote ICP ledger canister when deployed on mainnet.

I am happy to approve if majority disagrees with that.

PR #1191, opened 2025-06-18
standardize README

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/motoko/who_am_i/README.md
+++ b/motoko/who_am_i/README.md
@@ -16,7 +16,11 @@ When viewing this project in ICP Ninja, you can deploy it directly to the mainne
 
 ### 3. Navigate into the project's directory.
 
-### 4. Run `dfx start --background --clean && dfx deploy` to deploy the project to your local environment. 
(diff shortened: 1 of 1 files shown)


motoko/who_am_i/README.md
```diff
@@ -2,22 +2,22 @@
 
 Who am I? demonstrates how entities on the Internet Computer are identified. Every entity, such as a user or canister smart contract, has a principal identifier. Principals can be used for identification and authentication. Who am I? uses Internet Identity (II) for user authentication, then displays the principal identifier associated with that Internet Identity on the user interface.
 
-This application's logic is written in [Motoko](https://internetcomputer.org/docs/motoko/main/getting-started/motoko-introduction), a programming language designed specifically for developing canisters on ICP.
```
fxgst, line 5: This is kept in some other places

jessiemongeon1: I've removed it from the PR that had it included.

PR #1195, opened 2025-06-18

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


maksymar:
this example requires access to management canister for creating/loading canister snapshots, which ICP-ninja currently does not support.
I would recommend to keep all the current manual instructions in the terminal until ICP-ninja can fully support running such an example.

maksymar:
my comment above still holds, it's not possible to use this example in ICP-ninja, so I recommend reverting the change and keeping manual instructions.

PR #1196, opened 2025-06-18
add: Deploy to ICP Ninja + Standardize README

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/rust/composite_query/README.md
+++ b/rust/composite_query/README.md
@@ -1,62 +1,21 @@
 # Composite queries
 
-## Prerequisites
-This example requires an installation of:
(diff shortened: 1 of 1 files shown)


maksymar:
1) it should be possible to run this example fully on ICP-ninja, but nevertheless I would suggest keeping the manual instructions for users who want to try it locally in terminal and provide an extra link to ICP-ninja. would it be possible?

2) I was not able to run this particular example on ICP-ninja due to some build errors.

PR #1206, opened 2025-06-18

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


maksymar:
this example requires access to management canister to change canister settings, which is not supported by ICP-ninja at the moment.

PR #1212, opened 2025-06-18

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/rust/simd/README.md
+++ b/rust/simd/README.md
@@ -27,8 +27,34 @@ The example consists of a canister named `mat_mat_mul` (matrix-matrix multiplica
 dfx start --background --clean && dfx deploy
 ```
 
-Compare the amount of instructions used for different matrix multiplication implementations. Call a loop performing 1K element-wise multiplications of `K x 4` packed slices
-from matrices `A` and `B` using optimized algorithm, the same algorithm with Rust auto-vectorization enabled, and WebAssembly SIMD instructions:
+### 5. Open another terminal window in the same directory.
+
+```sh
+cd examples/rust/simd
+```
+
+### 6. Compile and deploy mat_mat_mul canister.
+
+```sh
+dfx deploy
+```
+
+Example output:
+
+```sh
+% dfx deploy
+[...]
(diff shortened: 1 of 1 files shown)


maksymar:
this example is working in ICP-ninja, but I would like to ask if it'd be possible to keep the manual instructions for advanced usage and add ICP-ninja link on top of that for quick start?

rust/simd/README.md
```diff
@@ -149,53 +102,6 @@ Rust auto-vectorization again demonstrates its power in this example.
 The auto-vectorized version of the integer matrix multiplication achieves
 more than a 2x speedup compared to the original code.
 
-## Further learning
-
-1. Have a look at the locally running dashboard. The URL is at the end of the `dfx start` command: `Dashboard: http://localhost/...`
-2. Check out `mat_mat_mul` canister Candid user interface. The URLs are at the end of the `dfx deploy` command: `mat_mat_mul: http://127.0.0.1/?canisterId=...`
-
-### Canister interface
```
maksymar, line 157: I would keep `Further learning` and `Conclusion` section to cover specific methods related to Wasm SIMD instructions.

rust/simd/README.md
```diff
@@ -149,53 +102,6 @@ Rust auto-vectorization again demonstrates its power in this example.
...
-  1K element-wise multiplications of `K x 4` packed slices
-  from matrices `A` and `B` using WebAssembly SIMD instructions.
-
-Example usage:
-
-```sh
-dfx canister call mat_mat_mul naive_f32
-```
-
-## Conclusion
```
maksymar, line 184: ditto

rust/simd/README.md
jessiemongeon1: IMO neither section's content includes anything meaningful. `Further learning` includes basic dfx start and Candid information, which is not specific to Wasm SIMD, and `Conclusion` reiterates things covered previously.

The `Canister interface` subcategory could be restored, but I would not keep either of the others.

PR #1213, opened 2025-06-18
add: Deploy to ICP Ninja + Standardize README

jessiemongeon1:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/rust/threshold-ecdsa/README.md
+++ b/rust/threshold-ecdsa/README.md
@@ -14,6 +14,19 @@ More specifically:
 
 [![](https://icp.ninja/assets/open.svg)](https://icp.ninja/editor?g=https://github.com/dfinity/examples/tree/master/rust/threshold-ecdsa)
 
+### 1. Update source code with the right key ID.
+
+To deploy the sample code on ICP Ninja, the canister needs the right key ID for the right environment. Specifically, one needs to replace the value of the `key_id` in the `src/ecdsa_example_rust/src/lib.rs` file of the sample code. Before deploying to the mainnet from ICP Ninja, one should modify the code to use the right name of the `key_id`.
+
+There are three options:
+
+* `dfx_test_key`: a default key ID that is used in deploying to a local version of IC (via IC SDK).
+* `test_key_1`: a master **test** key ID that is used in mainnet.
+* `key_1`: a master **production** key ID that is used in mainnet.
+
+> [!WARNING]
+> To deploy to IC mainnet, one needs to replace the value in `key_id `fields with the values `EcdsaKeyIds::TestKeyLocalDevelopment.to_key_id()` (mapping to `dfx_test_key`) to instead have either `EcdsaKeyIds::TestKey1.to_key_id()` (mapping to `test_key_1`) or `EcdsaKeyIds::ProductionKey1.to_key_id()` (mapping to `key_1`) depending on the desired intent.
+
 ## Build and deploy from the command-line
 
 ### 1. [Download and install the IC SDK.](https://internetcomputer.org/docs/building-apps/getting-started/install)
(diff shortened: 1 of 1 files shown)


rust/threshold-ecdsa/README.md
```diff
@@ -10,59 +10,25 @@ More specifically:
...
-
-- Acquire cycles (the equivalent of "gas" on other blockchains). This is necessary for all canisters.
-- Update the sample source code to have the right key ID. This is unique to this canister.
-
-### Acquire cycles to deploy
-
-Deploying to the Internet Computer requires [cycles](https://internetcomputer.org/docs/current/developer-docs/getting-started/tokens-and-cycles) (the equivalent of "gas" on other blockchains).
-
-### Update source code with the right key ID
+### 5. Update source code with the right key ID.
```
fspreiss: Before this PR, we had the two deployment options `Local deployment` and `Deploying the canister on the mainnet`, and the step `Update source code with the right key ID` was a sub-heading of the latter (as it is only needed there).

Do I understand correctly that the intention of this PR is to
* Replace `Deploying the canister on the mainnet` with `Deploying from ICP Ninja`, so as to make deploying to mainnet easier by means of using ICP Ninja
* Rename `Local Deployment` to `Build and deploy from the command-line` and simplify the section so that things boil down to a single command (namely `dfx start --background --clean && dfx deploy`).

If so, then the step `Update source code with the right key ID` should be a sub-heading of `Deploying to ICP Ninja`

jessiemongeon1: Thanks, @fspreiss ! I've adjusted accordingly.

rust/threshold-ecdsa/README.md
```diff
@@ -10,61 +10,13 @@ More specifically:
 - The sample canister hashes the message and uses the key derivation string for the derivation path.
 - The sample canister uses the above to request a signature from the threshold ECDSA [subnet](https://wiki.internetcomputer.org/wiki/Subnet_blockchain) (the threshold ECDSA is a subnet specializing in generating threshold ECDSA signatures).
 
-This tutorial gives a complete overview of the development, starting with downloading the [IC SDK](https://internetcomputer.org/docs/current/developer-docs/setup/index.md), up to the deployment and trying out the code on the IC mainnet.
+## Deploying from ICP Ninja
 
-> [!TIP]
```
altkdf, line 15: Is that a new convention to not include cross-references in examples in general?

rust/threshold-ecdsa/README.md
```diff
@@ -10,61 +10,13 @@ More specifically:
...
-> This walkthrough focuses on the version of the sample canister code written in the Rust programming language. There is also a [Motoko](https://internetcomputer.org/docs/current/developer-docs/backend/motoko/index.md) version available in the same repo and follows the same commands for deploying.
+[![](https://icp.ninja/assets/open.svg)](https://icp.ninja/editor?g=https://github.com/dfinity/examples/tree/master/rust/threshold-ecdsa)
 
-## Prerequisites
-This example requires an installation of:
+### 1. Update source code with the right key ID.
```
fspreiss, line 17: Currently the `Depoy from ICP Ninja` section has an implicit Step 0 (which is to click on the Button "Open in ICP Ninja") and this Step 1 here. But then the user isn't finished. In ICP Ninja, he still needs to click "Run canisters" for the application to be deployed, so I'm wondering if we should consider making this additional step explicit here.

@altkdf, for the future (not in this PR) we could consider making deploying to ICP Ninja even more seamless, e.g., be directly using a key that can be used on mainnet so that the Step 1 disappears entirely.

rust/threshold-ecdsa/README.md
jessiemongeon1: Yes

PR #1225, opened 2025-06-24
fix: npm prebuild scripts

jessiemongeon1:
Fixes build scripts for examples to deploy in ICP Ninja.

### DIFF
--- a/motoko/vetkd/package.json
+++ b/motoko/vetkd/package.json
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "build": "webpack",
-    "prebuild": "dfx generate",
+    "prebuild": "npm i --include=dev && npm run generate && dfx generate",
     "start": "webpack serve --mode development --env development",
     "deploy:local": "dfx deploy --network=local",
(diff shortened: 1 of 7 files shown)


lmuntaner (approved):
LGTM

sesi200 (approved):

motoko/vetkd/package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "build": "webpack",
-    "prebuild": "dfx generate",
+    "prebuild": "npm i --include=dev && npm run generate && dfx generate",
```
fspreiss, line 13: Looks generally good. I'm wondering though what the motivation for this change is: maybe add a brief PR description explaining why this is done (e.g., is this related to making things ready for ICP Ninja?). Also, do we need the same change then in the respective Rust example(s), e.g., `rust/vetkd/package.json`?

jessiemongeon1: Yes, this is for Ninja.

PR #1226, opened 2025-06-26
chore: Use parentheticals to attach cycles

alexandru-uta:
This PR updates the motoko http outcall examples to use the parenthetical notation instead of the deprecated way to attach cycles.

### DIFF
--- a/motoko/send_http_get/src/send_http_get_backend/main.mo
+++ b/motoko/send_http_get/src/send_http_get_backend/main.mo
@@ -1,5 +1,4 @@
 import Blob "mo:base/Blob";
-import Cycles "mo:base/ExperimentalCycles";
 import Nat64 "mo:base/Nat64";
 import Text "mo:base/Text";
 import IC "ic:aaaaa-aa";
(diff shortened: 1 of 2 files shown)


alexandru-uta:
> We could also have a separate example (or as another alternative example here) that uses https://mops.one/ic/docs/Call#httpRequest

If someone helps setting up mops with the CI jobs, I'm happy to add that.

Kamirus:
**EDIT**: Actually `moc=0.14.9` is currently bundled with dfx yes? But we would need 0.14.10 for this... So let's address this after dfx has fresher moc

Maybe it's easy to follow what is already in this repo:
- setting up in the workflow: https://github.com/dfinity/examples/blob/cb4fc3f0ccd3ad641f5e8b12fc093f362f174248/.github/workflows/motoko-basic-bitcoin.yaml#L25
- Readme update: https://github.com/dfinity/examples/blob/cb4fc3f0ccd3ad641f5e8b12fc093f362f174248/motoko/basic_bitcoin/README.md?plain=1#L33
- `mops.toml` example, moc must be 0.14.10 https://github.com/dfinity/examples/blob/cb4fc3f0ccd3ad641f5e8b12fc093f362f174248/motoko/filevault/mops.toml#L3-L5
- https://github.com/search?q=repo%3Adfinity%2Fexamples%20mops&type=code

PR #1234, opened 2025-07-03
doc(basic_bitcoin): add icp ninja link

mducroux:
Changes to the `basic_bitcoin` example `README.md`:
- Add a link to ICP ninja
- Remove reference to pre-deployed canister 
- Fix invalid candid args in `get_block_headers` command
- Fix invalid regtest address

### DIFF
--- a/rust/basic_bitcoin/README.md
+++ b/rust/basic_bitcoin/README.md
@@ -140,7 +140,7 @@ Example:
 
 ```bash
 dfx canister call basic_bitcoin send_from_p2pkh_address '(record {
-  destination_address = "tb1ql7w62elx9ucw4pj5lgw4l028hmuw80sndtntxt";
+  destination_address = "bcrt1qg8qknn6f3txqg97gt8ca0ctya0vw7ep6d02qmt";
   amount_in_satoshi = 4321;
 })'
 ```
@@ -159,9 +159,9 @@ The function returns the transaction ID. When interacting with the contract depl
 You can query historical block headers:
 
 ```bash
-dfx canister call basic_bitcoin get_block_headers '(10: nat32)'
+dfx canister call basic_bitcoin get_block_headers '(10: nat32, null)'
 # or a range:
-dfx canister call basic_bitcoin get_block_headers '(0: nat32, 11: nat32)'
+dfx canister call basic_bitcoin get_block_headers '(10: nat32, opt (11: nat32))'
 ```
 
 This calls `bitcoin_get_block_headers`, which is useful for blockchain validation or light client logic.
--- a/rust/basic_bitcoin/README.md
+++ b/rust/basic_bitcoin/README.md
@@ -38,7 +38,13 @@ For background on the ICP<>BTC integration, refer to the [Learn Hub](https://lea
 - [x] [Local Bitcoin testnet (regtest)](https://internetcomputer.org/docs/build-on-btc/btc-dev-env#create-a-local-bitcoin-testnet-regtest-with-bitcoind)
 - [x] On macOS, an `llvm` version that supports the `wasm32-unknown-unknown` target is required. This is because the Rust `bitcoin` library relies on the `secp256k1-sys` crate, which requires `llvm` to build. The default `llvm` version provided by XCode does not meet this requirement. Instead, install the [Homebrew version](https://formulae.brew.sh/formula/llvm), using `brew install llvm`.
 
-## Building and deploying the smart contract
+## Deploying from ICP Ninja
+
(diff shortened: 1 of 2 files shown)


rust/basic_bitcoin/README.md
```diff
@@ -38,7 +38,13 @@ For background on the ICP<>BTC integration, refer to the [Learn Hub](https://lea
 - [x] [Local Bitcoin testnet (regtest)](https://internetcomputer.org/docs/build-on-btc/btc-dev-env#create-a-local-bitcoin-testnet-regtest-with-bitcoind)
 - [x] On macOS, an `llvm` version that supports the `wasm32-unknown-unknown` target is required. This is because the Rust `bitcoin` library relies on the `secp256k1-sys` crate, which requires `llvm` to build. The default `llvm` version provided by XCode does not meet this requirement. Instead, install the [Homebrew version](https://formulae.brew.sh/formula/llvm), using `brew install llvm`.
 
-## Building and deploying the smart contract
+## Deploying from ICP Ninja
+
+This example can be deployed directly to the Internet Computer using ICP Ninja, where it will connect to Bitcoin **testnet4**.
```
lpahlavi: nit: I would add a note for what needs to be done to adapt `dfx.json` so that the example works locally after downloading it from ICP Ninja (i.e. change `testnet` to `regtest`).

lpahlavi (approved):

rust/basic_bitcoin/README.md
mducroux: If users follow the tutorial, at step 4 they'll encounter the command: `dfx deploy basic_bitcoin --argument '(variant { regtest })'` which deploys the example to `regtest` (command line arguments have priority over dfx.json args). So there is no need to modify the `dfx.json` file. Do you think I should still add it here?

rust/basic_bitcoin/README.md
```diff
@@ -38,7 +38,13 @@ For background on the ICP<>BTC integration, refer to the [Learn Hub](https://lea
...
+
+[![](https://icp.ninja/assets/open.svg)](https://icp.ninja/editor?g=https://github.com/dfinity/examples/tree/master/rust/basic_bitcoin)
```
mducroux, line 45: This points to the current master which does not include the latest change on dfx.json (namely adding init_arg). For the purpose of this review, dfx.json on ICP ninja needs to be modified to include init_arg.

rust/basic_bitcoin/dfx.json
```diff
@@ -14,7 +14,8 @@
       ],
       "package": "basic_bitcoin",
       "type": "custom",
-      "wasm": "target/wasm32-unknown-unknown/release/basic_bitcoin.wasm"
+      "wasm": "target/wasm32-unknown-unknown/release/basic_bitcoin.wasm",
+      "init_arg": "(variant { testnet })"
```
kristoferlund, line 18: If `dfx.json` specifies a specific network, then the README should clearly describe this. Currently, you are instructed to start a regtest network but the default `dfx.json` config uses testnet. Make sure the README matches the default setup. 

Consider making `regtest` the default setting for `dfx.json` and creating a separate branch for the Ninja version. That branch then can use `testnet`.

mducroux: I can surely add some precision in the README that testnet is the default network. However, I think that having a separate branch will make it harder to maintain the example, i.e. whenever we update the example, we will need to update both the master branch and the ninja branch which is not ideal. WDYT?

kristoferlund: Yeah, that would have made sense if running on Ninja is the main intended use. It isn't. The main intended use for the example is for a dev to fork it and use as a starting point for their own development. Testrunning on Ninja is usecase number 2. 

If you make testnet the default network, included in `dfx.json`, you solve one problem (the Ninja deploy problem) and create a new problem (the local deploy problem). We should instead attempt to solve both problems.

kristoferlund: Ideally, there would be a `.ninja-config.json` or similar, where you can specify Ninja specific things such as default `init_args`.

mducroux: I don't think it creates a problem as long as the README explicitly says that deploying the example should be done using `dfx deploy basic_bitcoin --argument '(variant { regtest })'` or do you see a bigger problem?

kristoferlund: Does that override the `dfx.json` setting? In that case then both problems are solved.

mducroux: Yes it does.

PR #1237, opened 2025-07-10
chore: Single-replica http outcalls update motoko example

alexandru-uta:
This PR updates the http outcall motoko examples to use single-replica http outcalls.


raymondk (approved):

letmejustputthishere:
should we update the rust examples as well?

PR #1244, opened 2025-08-01
Update Motoko examples to use new standard library

rvanasa:
Work in progress!

This PR updates all Motoko examples which use the original `base` library to use the new `core` package instead.

* [Documentation](https://internetcomputer.org/docs/motoko/core)
* [Migration guide](https://internetcomputer.org/docs/motoko/base-core-migration)


rvanasa:
Closing as a result of diverging from the main branch (diff can be used as reference for updating the examples in the future).

PR #1245, opened 2025-08-05
Feature/add-motoko-nft-creator

letmejustputthishere:
**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


motoko/nft-creator/frontend/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "frontend",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "prebuild": "dfx generate",
```
fxgst: Looks like packages are not installed thats why CI is failing (would also fail on Ninja)

```suggestion
    "prebuild": "npm i --include=dev && dfx generate",
```

fxgst: PR to include it in the guide: https://github.com/dfinity/examples/pull/1246

StepankaHru (approved):

PR #1252, opened 2025-08-14
chore: delete old examples

sesi200:


ggreif:
You could make an `attic/` folder and move stuff there.

PR #1253, opened 2025-08-14
Add `core` dependency to examples

rvanasa:
This PR makes it possible to use the new Motoko standard library (`core`) package in parallel with `base`, primarily for ICP Ninja.

We eventually want to convert these examples to entirely use `core` (see https://github.com/dfinity/examples/pull/1244).


letmejustputthishere (approved):

fxgst (approved):
Thanks! You could also add it to motoko/nft-creator, then we'd cover all Motoko examples on ninja from the examples repo. You can see [here](https://github.com/dfinity/examples/blob/cfbef548b88d3693cde79412dadaf4e28150a97c/.github/workflows/ninja_pr_checks.yml) which examples show up in ninja

THLO (approved):

PR #1259, opened 2025-08-26
chore: prepare more examples for Ninja

sesi200:


sesi200:
The removed `Makefile`s did not test anything beyond `dfx deploy` working, which the Ninja workflow does already.

The package-lock files are not very useful because they get outdated all the time, and we don't run `npm build` with locked pages.json anyways.

Yes, the removed workflows would just duplicate work

berestovskyy (approved):

PR #1261, opened 2025-08-28
Add an http outcalls example

mihailjianu1:
This showcases two ways of using the http outcalls feature in rust, and adds them to the icp ninja examples:
- A replicated http outcall to a constant, fairly deterministic service
- a non replicated outcall to a volatile service

In the case of the replicated http outcall, we also showcase the use of transform queries and how / why they help reach consensus.


rust/http_outcalls/backend/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        url: "https://api.coingecko.com/api/v3/simple/price?ids=bitcoin&vs_currencies=usd".to_string(),
+        // The maximum size (in bytes) of the response before, or after transformation.
+        // If "None" is used, it defaults to 2MB.
+        max_response_bytes: Some(10000),
+        method: HttpMethod::GET,
+        headers: vec![],
+        body: None,
+        // No need to transform the response any longer, as it is not replicated
+        transform: None,
+        // The request is not replicated. This means that a single node will attempt the outcall. This is ideal for fast moving data (such as eg crypto prices), non idempotent operations and so on.
```
derlerd-dfinity: I think we should mention here that one must be cautious with trusting such a response, as a malicious replica could return whatever it wanted. And maybe also that for practical purposes this usually means that one needs a way to verify the response out of band inside the canister, or to send out multiple such requests and only accept if enough responses agree.

derlerd-dfinity: Similarly, I'd also mention something along those lines in the readme.md.

mihailjianu1: Good points, I added some more comments both in the code and in the readme doc

PR #1263, opened 2025-08-28
chore: prepare more examples for Ninja

sesi200:


berestovskyy (approved):

georgi-dfinity (approved):

motoko/threshold-schnorr/package-lock.json
fspreiss, line 1: Are the package-lock files no longer useful?

motoko/threshold-schnorr/test.sh
```diff
@@ -21,6 +21,7 @@ function test_impl() {
     ed25519_public_key_hex="$(get_text_in_double_quotes "${ed25519_public_key_raw_output}")"
     echo ed25519_public_key_hex="$ed25519_public_key_hex"
 
+    npm install
```
fspreiss: Maybe put that in the Makefile instead?

motoko/threshold-ecdsa/BUILD.md
```diff
@@ -0,0 +1,113 @@
+# Continue building locally
+
+Projects deployed through ICP Ninja are temporary; they will only be live for 20 minutes before they are removed. The command-line tool `dfx` can be used to continue building your ICP Ninja project locally and deploy it to the mainnet.
+
+To migrate your ICP Ninja project off of the web browser and develop it locally, follow these steps.
```
fspreiss, line 5: The file/text assumes that you started on ICP Ninja, and then want to continue building locally.

But developers also could directly find this example by browsing github.com/dfinity/examples, and then read the BUILD.md: won't they be confused then, especially if they don't know what ICP Ninja is?

One way to solve this is by rewriting the text so that it fits for both scenarios (devs coming via ICP Ninja and devs coming directly to github.com/dfinity/examples.

motoko/threshold-schnorr/package-lock.json
sesi200: No, the frontend build has never respected package-lock, only Cargo.lock

viviveevee: In my experience not really since everything is outdated within minutes in the node ecosystem, and Ninja has rather strict project size limits, so I would rather keep the data volume out

motoko/threshold-ecdsa/BUILD.md
fxgst: This is consistent in all examples for ICP Ninja: the README.md contains general information, the BUILD.md is specific for ICP Ninja

viviveevee: This text is standard in _all_ Ninja-compatible examples. Let's discuss improvements to this in a separate PR

fspreiss: > Let's discuss improvements to this in a separate PR

👍 

> This is consistent in all examples for ICP Ninja: the README.md contains general information, the BUILD.md is specific for ICP Ninja

Having this consistent across all examples make sense. Still, if a developer finds the github.com/dfinity/examples repo and looks at an example and finds the BUILD.md file and reads it expecting infos regarding how to build the example, without even knowing that ICP Ninja exists, this developer will be confused when reading the main heading "Continue building locally".

motoko/threshold-schnorr/package-lock.json
fspreiss: I find it useful because it makes the example reproducible. Given that this particular package-lock.json only locks a few crypto libraries (which are used for testing) and is not particularly big (156 lines) my preference would be to keep it.

.github/workflows/rust-x509-example.yml
```diff
@@ -1,45 +0,0 @@
...
-      - name: Provision Linux
-        run: bash .github/workflows/provision-linux.sh
-      - name: Provision PocketIC
-        run: bash .github/workflows/provision-pocket-ic-server.sh
-      - name: Rust X.509 Linux
-        run: |
-          sudo apt-get install --yes openssl
-          pushd rust/x509
-          dfx build --check
-          cargo test
```
fspreiss: I only realized now that this openssl installation and the cargo testing are not covered by the new `ninja_pr_checks.yml` (as it only does `dfx start --background` and `dfx deploy`, and maybe `make test`).

So I guess we need a new Makefile here that does the openssl installation and `cargo test` as part of a test target.

mducroux (approved):

rust/basic_ethereum/README.md
```diff
@@ -14,7 +14,13 @@ features of the Internet Computer.
...
 the [Ethereum integration overview](https://internetcomputer.org/docs/current/developer-docs/multi-chain/ethereum/overview).
 
-## Prerequisites
+## Deploying from ICP Ninja
+
+When viewing this project in ICP Ninja, you can deploy it directly to the mainnet for free by clicking "Run" in the upper right corner. Open this project in ICP Ninja:
+
+[![](https://icp.ninja/assets/open.svg)](https://icp.ninja/i?g=https://github.com/dfinity/examples/rust/basic_ethereum)
+
+## Deploying locally
```
gregorydemay: This section is not about deploying locally, in fact it's about deploying on the Internet Computer. There are actually good reasons why one shouldn't deploy locally for anything using HTTPs outcall as it may give a false sense of things to be working when they would not on mainnet. I think it would generally be useful to have somewhere a strong disclaimer about the differences between deploying locally vs deploying on mainnet and link it here (if we want to add a section on deploying locally)

fxgst: You're right, let me revert that change and just add a section to deploy it from ICP Ninja

motoko/threshold-schnorr/package-lock.json
fxgst: Brought it back

.github/workflows/rust-x509-example.yml
fxgst: brought the separate workflow back

rust/guards/dfx.json
```diff
@@ -15,11 +15,5 @@
...
   },
-  "defaults": {
-    "build": {
-      "args": "",
-      "packtool": ""
-    }
-  },
   "version": 1
-}
+}
```
fxgst:
```suggestion
}

```

rust/qrcode/package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "build": "webpack",
-    "prebuild": "dfx generate",
+    "prebuild": "npm install --include=dev && dfx generate",
```
ilbertt, line 13: Did we add this because the `NODE_ENV` is set to `production` on ICP Ninja? ([NPM docs ref](https://docs.npmjs.com/cli/v11/commands/npm-install#omit)).

It is not common to have to run install on a prebuild, and the config should rather change on ICP Ninja in my opinion.

If we can't change it, a `preinstall`/`postinstall` [script](https://docs.npmjs.com/cli/v11/using-npm/scripts) may be more appropriate here.

svelte/svelte-starter/package.json
```diff
@@ -9,10 +9,13 @@
     "Canister",
     "Internet Identity"
   ],
+  "workspaces": [
+    "src/frontend"
+  ],
   "scripts": {
     "build": "cd src/frontend && npm run build",
-    "prebuild": "dfx generate",
+    "prebuild": "npm install --include=dev && dfx generate",
```
ilbertt, line 17: Same as my other comment

ilbertt (changes requested):

rust/qrcode/package.json
fxgst: we want to avoid additional scripts for building the canisters, they should just build with dfx build. This is the same across all ninja examples

svelte/svelte-starter/package.json
fxgst: See above

eichhorl (approved):

rust/qrcode/package.json
ilbertt: As discussed offline, this pattern is not standard and can create confusion for people that either copy the example and run it locally or that use it on ICP Ninja. ICP Ninja should manage dependencies without having to rely on the build step.

Anyway, since there are already other examples that follow this pattern, I'm going to approve this

ilbertt (approved):

motoko/threshold-schnorr/src/schnorr_example_motoko/main.mo
```diff
@@ -36,6 +36,7 @@ persistent actor {
   };
 
   transient var ic : IC = actor ("aaaaa-aa");
+  transient var key_id : Text = "test_key_1"; // Use "key_1" for production and "dfx_test_key" locally
```
fspreiss, line 39: Did you mean to also use this in lines 49 and 72? Currently this new variable is unused, no?

fspreiss: This was addressed in https://github.com/dfinity/examples/pull/1265.

PR #1286, opened 2025-10-15
feat: example on receiving ICP

ielashi:
Adds an example on how to receive ICP. This example will be featured in a tutorial in our defi docs where users can deploy this project to ICP.ninja, get TESTICP tokens from a faucet, then verify that they've received their tokens from their balance.

Note that this example doesn't have tests. I'd like to make the faucet pullable to make this project easy to run and test locally, which I'll address in future commits.

### DIFF
--- a/rust/receiving-icp/src/lib.rs
+++ b/rust/receiving-icp/src/lib.rs
@@ -5,31 +5,37 @@ use ic_ledger_types::{AccountBalanceArgs, AccountIdentifier, Subaccount};
 // To use real ICP, use `ryjl3-tyaaa-aaaaa-aaaba-cai` instead.
 const LEDGER_PRINCIPAL: &str = "xafvr-biaaa-aaaai-aql5q-cai";
 
+fn get_account(upper: u128, lower: u128) -> AccountIdentifier {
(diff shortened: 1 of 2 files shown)


letmejustputthishere:
you could (locally) also rely on this new feature shipping with dfx
https://github.com/dfinity/sdk/blob/master/CHANGELOG.md#feat-dfx-start---system-canisters-for-bootstrapping-system-canisters

ielashi:
> > Ledger is a completely independent unit from the NNS. Wanting to install a ledger != wanting to install the entirety of the NNS.
> 
> Not that it changes the other points you mentioned, but wanted to clarify that you don't need to install all canisters on the NNS, you can [cherry pick them](https://github.com/dfinity/ic/blob/dba88a31443d41f0a9b5d9db60e2d4d1d4c20c24/packages/pocket-ic/src/common/rest.rs#L583) to my understanding

Ah, noted, thank you. In this case, may I get editorial approval?

ielashi:
@marc0olo thanks for the comments! Like our documentation I think we can think of it iteratively. For now this example is meant to be very small in scope and we can add more as needed. Of course, PRs to improve it and the associated tutorial would be more than welcome :)

PR #1322, opened 2026-04-13
feat(basic_bitcoin): add get_blockchain_info example

mducroux:
- Upgrade `ic-cdk` to v0.20.0 in the basic_bitcoin example, migrating to the new standalone `ic-cdk-bitcoin-canister` and `ic-cdk-management-canister` crates and adapting to API changes (`NetworkInRequest`, typed `Txid`)                          
- Add `get_blockchain_info` endpoint that returns the Bitcoin chain tip height, block hash, timestamp, difficulty, and UTXO count


rust/basic_bitcoin/src/service/get_blockchain_info.rs
```diff
@@ -0,0 +1,14 @@
+use crate::BTC_CONTEXT;
+use ic_cdk::update;
+use ic_cdk_bitcoin_canister::BlockchainInfo;
+
+/// Returns information about the state of the Bitcoin blockchain
+/// including the tip height, block hash, timestamp, difficulty, and UTXO count.
```
THLO, line 6:
```suggestion
/// Returns information about the state of the Bitcoin blockchain:
/// The tip height, block hash, timestamp, difficulty, and UTXO count.
```

rust/basic_bitcoin/src/common.rs
```diff
@@ -112,7 +112,7 @@ pub fn build_transaction_with_fee(
         .iter()
         .map(|utxo| TxIn {
             previous_output: OutPoint {
-                txid: Txid::from_raw_hash(Hash::from_slice(&utxo.outpoint.txid).unwrap()),
+                txid: Txid::from_raw_hash(Hash::from_slice(utxo.outpoint.txid.as_ref()).unwrap()),
```
THLO, line 115: Why is this change needed?

THLO (approved):

rust/basic_bitcoin/src/common.rs
mducroux: This is needed because `txid` went from `Vec<u8>` to `Txid` in `ic-cdk` `v0.20.0`. `Hash::from_slice` expectsa `&[u8]`. `&` was enough to coerce `Vec<u8>` → `&[u8]`, but for the new `Txid` you need `.as_ref()` to extract the underlying bytes.

PR #1325, opened 2026-04-20
chore(motoko): migrate composite_query, parallel_calls, random_maze, classes, hello_cycles to mo:core 2.4.0

marc0olo:
## Summary

- Add \`mops.toml\` with \`core = "2.4.0"\` to all five examples; remove \`mo:base\` dependency
- Replace all \`mo:base\` imports with \`mo:core\` equivalents
- Apply modern Motoko style via \`mops check --fix\`: context dot notation (e.g. \`map.get(k)\`) and inferred implicit arguments (e.g. compare functions auto-resolved)
- Enable compiler warnings M0236, M0237, M0223 in \`mops.toml\` (with inline comments explaining each flag) to enforce the modern style going forward
- Update READMEs for \`random_maze\`, \`classes\`, and \`composite_query\` to reflect the new APIs
- Notable API migrations:
  - \`composite_query\` / \`classes\`: \`RBTree\` → \`Map\`, \`ExperimentalCycles.add(n); await expr\` → \`await (with cycles = n) expr\`, \`Array.init\` → \`VarArray.repeat\`
  - \`parallel_calls\`: functional \`mo:base/List\` → mutable \`mo:core/List\` (cleaner: no push+reverse needed), \`Iter.range\` → \`Nat.range\`
  - \`random_maze\`: \`Random.Finite\` + manual entropy blobs → \`Random.AsyncRandom\` via \`Random.crypto()\` (entropy fetched automatically), \`Stack\` class → \`Stack\` module functions, \`mo:base/List\` → \`mo:core/pure/List\`
  - \`hello_cycles\`: \`ExperimentalCycles\` → \`mo:core/Cycles\`, \`Cycles.add(n); await expr\` → \`await (with cycles = n) expr\`, \`Cycles.accept\` → \`Cycles.accept<system>\`

**Codeowner:** @dfinity/languages

## Test plan

- [ ] \`mops install\` succeeds in each example directory
- [ ] \`dfx build\` succeeds for all five examples
- [ ] \`random_maze\` generate function returns a valid maze string
- [ ] \`hello_cycles\` wallet functions work correctly with the new cycle-attachment syntax
- [ ] \`classes\` / \`composite_query\` map operations (get/put) work end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1327, opened 2026-04-20
chore: migrate low_wasm_memory to mo:core 2.4.0

marc0olo:
## Summary

Missed in #1325 (same @dfinity/languages batch).

- Add `mops.toml` with `moc = "1.5.1"` and annotated `[moc]` warning flags
- Replace `mo:base/Array`, `mo:base/Buffer`, `mo:base/Deque` with `mo:core/Array`, `mo:core/List`, `mo:core/Queue`
- Apply context dot notation (`fnOrderBuffer.add`, `fnOrderBuffer.toArray`, `bytes.pushBack`)
- Remove redundant type annotations

**Codeowner:** @dfinity/languages

### DIFF
--- a/motoko/low_wasm_memory/src/low_wasm_memory_hook/main.mo
+++ b/motoko/low_wasm_memory/src/low_wasm_memory_hook/main.mo
@@ -1,6 +1,6 @@
-import Array "mo:base/Array";
-import Deque "mo:base/Deque";
-import Buffer "mo:base/Buffer";
+import Array "mo:core/Array";
+import Queue "mo:core/Queue";
+import List "mo:core/List";
 
 persistent actor {
   // Types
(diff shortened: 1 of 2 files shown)


motoko/low_wasm_memory/src/low_wasm_memory_hook/main.mo
```diff
@@ -10,22 +10,22 @@ persistent actor {
   };
 
   // State
-  transient var fnOrderBuffer = Buffer.Buffer<FnType>(30);
-  transient var bytes : Deque.Deque<[Nat]> = Deque.empty();
+  transient let fnOrderBuffer = List.empty<FnType>();
+  transient let bytes = Queue.empty<[Nat]>();
   transient var hookExecuted : Bool = false;
```
Kamirus, line 15: Is this meant to be transient? The state should be reset on upgrade? If not then we can drop `transient` and these fields will be `stable` and persist over upgrades

Kamirus (approved):
LGTM, with a question:

motoko/low_wasm_memory/src/low_wasm_memory_hook/main.mo
marc0olo: I think it is just for demo'ing the hook and doesn't require to be persisted.

my guess is that this was intentionally marked `transient`. I see that you originally created the example and didn't define it as `stable` back then.

shall we merge?

Kamirus: Back then it was simpler to just add `transient`. That's ok let's keep transient, this PR is already an improvement

PR #1340, opened 2026-04-21
chore(motoko): migrate cert-var to mo:core 2.4.0

marc0olo:
## Summary
- Add `mops.toml`: moc 1.5.1, core 2.4.0, `-W=M0236,M0237,M0223`
- Update `dfx.json` packtool to `mops sources`
- Replace `mo:base` imports with `mo:core` equivalents
- Drop unused `Debug` import
- Use `x.toNat8()` context dot instead of `Nat8.fromNat(Nat32.toNat(x))`, dropping the `Nat8` import
- Fix M0194: `blobOfNat32` was using captured actor field `value` instead of its parameter `n` (functionally equivalent since it's always called with `value`, but causes an unused-parameter warning)

### DIFF
--- a/motoko/cert-var/src/cert_var/main.mo
+++ b/motoko/cert-var/src/cert_var/main.mo
@@ -1,10 +1,8 @@
-/// Simple counter (see `Counter.mo`), but uses `mo:base/CertifiedData` to
+/// Simple counter (see `Counter.mo`), but uses `mo:core/CertifiedData` to
 /// implement the counter value as a certified variable.
-import CD "mo:base/CertifiedData";
-import Blob "mo:base/Blob";
-import Nat8 "mo:base/Nat8";
-import Nat32 "mo:base/Nat32";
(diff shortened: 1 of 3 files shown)


motoko/cert-var/src/cert_var/main.mo
```diff
@@ -14,14 +12,12 @@ persistent actor Variable {
   /// LE encoding matches Candid encoding of Nat32, for consistency and convenience.
   func blobOfNat32(n : Nat32) : Blob {
     let byteMask : Nat32 = 0xff;
-    func byte(x : Nat32) : Nat8 {
-      Nat8.fromNat(Nat32.toNat(x));
-    };
+    func byte(x : Nat32) : Nat8 = x.toNat8();
```
raymondk, line 15: If this is really necessary maybe it should be in the certified data. It doesn't make sense that a user would have to implement it themselves.

raymondk (approved):

Issue #1347, opened 2026-05-22
Examples audit: restructure for docs integration (educational assessment)


marc0olo:
This issue captures a full educational audit of every example in this repository, conducted to determine what to keep, refocus, or remove in preparation for referencing examples from [dfinity/developer-docs](https://github.com/dfinity/developer-docs).

Every verdict is based on reading the actual source code, not just READMEs. The focus is educational and structural value only. Tooling migrations (dfx → icp-cli) are tracked separately.

The docs integration mapping at the bottom supersedes the aspirational mapping in [developer-docs#44](https://github.com/dfinity/developer-docs/issues/44) and can be used to drive Phase 2 of that work (adding `// #region` markers).

---

## Quick-reference verdict table

| Example | Verdict |
|---------|---------|
| `hosting/godot-html5-template` | Keep |
| `hosting/my_crypto_blog` | **Remove** |
| `hosting/oisy-signer-demo` | Keep |
| `hosting/photo-storage` | Keep but refocus |
| `hosting/react` | Keep |
| `hosting/static-website` | Keep |
| `hosting/unity-webgl-template` | Keep |
| `native-apps/unity_ii_deeplink` | Keep but refocus (consolidate) |
| `native-apps/unity_ii_applink` | **Remove** (merge into deeplink) |
| `native-apps/unity_ii_universallink` | **Remove** (merge into deeplink) |
| `svelte/svelte-motoko-starter` | Keep but refocus |
| `svelte/sveltekit-starter` | Keep but refocus |
| `wasm/counter` | Keep |
| `motoko/backend_only` | Keep but refocus |
| `motoko/basic_bitcoin` | Keep |
| `motoko/canister_factory` | Keep |
| `motoko/canister_logs` | Keep |
| `motoko/cert-var` | Keep but refocus |
| `motoko/classes` | Keep |
| `motoko/composite_query` | Keep |
| `motoko/daily_planner` | Keep but refocus |
| `motoko/encrypted-notes-dapp-vetkd` | **Remove** |
| `motoko/evm_block_explorer` | Keep |
| `motoko/filevault` | Keep |
| `motoko/flying_ninja` | Keep |
| `motoko/hello_cycles` | Keep |
| `motoko/hello_world` | Keep |
| `motoko/ic-pos` | Keep but refocus |
| `motoko/icp_transfer` | Keep but refocus |
| `motoko/icrc2-swap` | Keep |
| `motoko/llm_chatbot` | Keep |
[...]

PR #1348, opened 2026-05-25
chore: remove examples marked for removal

raymondk:
Removed all the examples marked for removal in: https://github.com/dfinity/examples/issues/1347

except:

- native-apps/unity_ii_applink
- native-apps/unity_ii_universallink

Which need to be merged into deeplink


mbjorkqvist (approved):
Thanks for the cleanup!

marc0olo (approved):
just a sidenote:
- for the icrc-token related examples we will create a new example that covers everything needed with region markers so that developer-docs can easily pick up important snippets

rust/basic_dogecoin/README.md
```diff
@@ -1,3 +0,0 @@
-# Basic Dogecoin
```
gregorydemay, line 1: @marc0olo It's fine by me to remove those (like for basic_solana), the question is do we need a replacement? The reason for having those files in the first place was:

1. it makes more sense to have those examples in their respective repository where the code lives
2. There was the ask to still keep one central place where dev could find those examples (the examples repo) and that's why there are those READMES pointing to where one could find the examples

marc0olo: 1) I guess that is fine, but it also means we most likely won't include specific snippets in the dev docs guides going forward, at least in the beginning.

2) for observability we can do following things:
- link to them in a example index here in the repo (instead of having a dedicated folder + readme)
- link to them in the respective guides in the docs. this is what we already do on these pages:
  - https://docs.internetcomputer.org/guides/chain-fusion/solana
  - https://docs.internetcomputer.org/guides/chain-fusion/dogecoin

btw, if you identify false claims on any of these docs, please let me know. I am happy to fix this asap then.

alin-at-dfinity (approved):

PR #1359, opened 2026-06-09
Adds a Blazor WebAssembly (.NET 10) + Motoko Hello World template.

mitnick2012:
**Overview**
The icp-cli templates library has no support for .NET/C# developers who want to 
build on ICP using Blazor WebAssembly. This template fills that gap by providing 
a working starter project that combines Blazor WASM (.NET 10) on the frontend with 
a Motoko canister on the backend, using icp-cli for deployment. Without this template,
.NET developers face undocumented pitfalls that make getting started unnecessarily 
difficult.

**Requirements**
- Blazor WASM frontend that successfully calls a Motoko backend canister
- Uses icp-cli with canister.yaml configuration (not dfx)
- Works on local network and mainnet
- Documents all known gotchas specific to Blazor + ICP
- No JS interop race conditions between @dfinity/agent and Blazor runtime
- Canister ID discovery via ic_env cookie without hardcoding

**Considered Solutions**
1. ES module script (`type="module"`) for @dfinity/agent — rejected due to browser 
   deferral behavior causing a race condition where Blazor attempts JS interop before 
   the module is ready, silently breaking button events.
2. Webpack bundle loaded with `defer` — selected. Compiles @dfinity/agent into a 
   single plain JS file exposed as `window.IcpAgent`. The `defer` attribute guarantees 
   script execution order before DOMContentLoaded, so the global is always available 
   when Blazor makes its first JS interop call.

**Recommended Solution**
Webpack + `defer` loading pattern, modeled after the existing `Dfinity.Blazor` 
community library. This is the only approach that reliably solves the JS interop 
race condition without adding complexity to the Blazor startup sequence. It also 
bundles @dfinity/agent locally (no CDN dependency), pins the version in package.json, 
and produces a single minified file that the asset canister can certify correctly.

**Considerations**
- **Security:** No impact. The template follows the same trust model as other icp-cli 
  templates — no private keys, no hardcoded canister IDs.
- **Performance:** The webpack bundle adds ~271KB (uncompressed) to the frontend assets. 
  This is negligible compared to the Blazor WASM runtime (~5MB+). Brotli compression 
  via the asset canister reduces the bundle to ~80KB on the wire.
- **Users:** Purely additive. Existing templates are unaffected. .NET developers gain 
  a working starting point without having to discover undocumented pitfalls.
- **Breaking changes:** None.
- **Maintenance:** Low. The template depends on @dfinity/agent and the @dfinity/motoko 
  recipe — both of which are maintained by DFINITY. The .NET 10 LTS support window 
  runs until November 2028, so no forced upgrade in

### DIFF
--- a/blazor/hello_world/.gitignore
+++ b/blazor/hello_world/.gitignore
@@ -1,0 +1,20 @@
+# ICP - cache only, keep .icp/data/ for canister IDs
+.icp/cache/
+
+# .NET build output
+bin/
+obj/
+
+# Node
+node_modules/
+
+# webpack generated output (regenerated on build)
+frontend/BlazorFrontend/wwwroot/icpAgent.js
+
+# OS
+.DS_Store
+Thumbs.db
+
+# Logs
+*.log
+npm-debug.log*
--- a/blazor/hello_world/README.md
+++ b/blazor/hello_world/README.md
@@ -1,0 +1,165 @@
+# ICP Blazor Hello World
+
+A hello world template combining **Blazor WebAssembly (.NET 10)** on the frontend and **Motoko** on the backend, deployed fully on-chain on the **Internet Computer (ICP)**.
+
+> The first working template for Blazor WASM + icp-cli + Motoko.
+
+## Stack
+
+| Layer    | Technology                          |
+|----------|-------------------------------------|
+| Frontend | Blazor WebAssembly (.NET 10)        |
+| Backend  | Motoko canister                     |
+| Bridge   | @dfinity/agent (webpack bundle)     |
+| Platform | Internet Computer (ICP)             |
+| CLI      | icp-cli                             |
+
+## Project Structure
+
+```
+icp-blazor-hello/
+├── icp.yaml                          # icp-cli project config
+├── .gitignore
+├── backend/
+│   ├── canister.yaml                 # backend canister config
+│   ├── backend.did                   # Candid interface (semicolons required!)
+│   └── src/
+│       └── main.mo                   # Motoko canister
+└── frontend/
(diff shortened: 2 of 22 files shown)


mitnick2012:
Hello @marc0olo,
Could you please review this feature?

marc0olo:
Hey @mitnick2012,

I will discuss with the team how to handle this. But I don't think we want to maintain external examples here in this repo.

In any case this would be better suited for the https://github.com/dfinity/icp-cli-templates repository. See this guide on how to create and use custom templates: https://cli.internetcomputer.org/1.0/guides/creating-templates

What I could imagine is that we provide an overview for community created templates either in the official icp-cli documentation or in the README of the icp-cli-templates repository.

I also noticed that you depend on `@dfinity/agent` in your example. You should migrate this to https://js.icp.build/core/latest/libs/agent

Also note that the latest Motoko recipe (`v5.0.0`) delegates the compilation entirely to `mops build`:
- https://github.com/dfinity/icp-cli-recipes/releases/tag/motoko-v5.0.0
- https://github.com/dfinity/icp-cli-templates/blob/main/hello-world/motoko-backend/mops.toml

Best,
Marco

mitnick2012:
Thanks @marc0olo  for the review I will update the MR with your remarks, and move it to template repo, my aim is to give the blazor users a starting push to build with icp

marc0olo:
@mitnick2012 for now my recommendation is following:
- create a "real" template as per guide above
- create a forum post about this template, share its github repo and advertise the command how to use it

independent of that I will think about how we can provide a registry for independent templates created by the community.

PR #1372, opened 2026-06-10
chore: migrate motoko/basic_bitcoin to icp-cli

marc0olo:
## Summary

Migrates `motoko/basic_bitcoin` from dfx to icp-cli with a self-contained local Bitcoin testing environment.

**Infrastructure:**
- Replace `dfx.json` with `icp.yaml` using `@dfinity/motoko@v5.0.0`; three environments: `local` (regtest), `staging` (testnet4), `production` (mainnet)
- `actor class BasicBitcoin(network : Types.Network)` configured via `init_args` per environment
- Custom Docker image (`icp-cli-network-launcher-bitcoin`) that extends the official network launcher with `bitcoind` in regtest mode — fully self-contained local testing, no external Bitcoin node required
- `bitcoind` RPC bound to `127.0.0.1` (all callers run inside the container)
- `make build-image` uses `--pull` to always fetch the latest base image, preventing stale local caches from causing issues
- `make test` fails fast with a clear error if the network launcher container is not running
- CI: runs on host runner (not inside a container) so icp-cli can bind-mount the Docker status dir; installs icp-cli + ic-wasm + ic-mops via npm; builds image with GHA layer cache
- Delete `dfx.json`, `BUILD.md`, `.devcontainer/`

**Canister code:**
- Move all source from `src/basic_bitcoin/src/` → `backend/`; rename `Main.mo` → `app.mo`
- `BitcoinApi.mo`: calls the Bitcoin canister **directly** at its network-specific principal (not the management canister proxy). `bitcoinCanister(network)` selects `ghsi2-...` (mainnet) or `g4xu7-...` (testnet/regtest) at runtime. Actor type mirrors the [official Bitcoin canister Candid](https://github.com/dfinity/bitcoin-canister/blob/master/canister/candid.did). All Bitcoin API calls — including `get_block_headers` — go through `BitcoinApi`
- `EcdsaApi.mo` / `SchnorrApi.mo`: use `mo:ic` directly (`ic.ecdsa_public_key`, `ic.sign_with_ecdsa`, `ic.schnorr_public_key`, `ic.sign_with_schnorr`) — threshold signing still goes through the management canister `aaaaa-aa`
- `KEY_NAME` derived from network: `key_1` (mainnet) or `test_key_1` (testnet4 + local regtest)
- Add `get_block_headers` and `get_blockchain_info` via `BitcoinApi`
- Update `mops.toml`: moc 1.9.0, core 2.5.0, ic 4.0.0, `--default-persistent-actors`
- `Types.mo`: reduced to types not available in `mo:bitcoin` or `mo:ic`; `Utxo`/`OutPoint` aliased from `mo:bitcoin`; use `Blob` for block hashes (not `[Nat8]`)

**Tests (`Makefile`):**
- 6 tests: address generation (P2PKH, P2TR key-only, P2TR full), fee percentiles, balance after mining, UTXOs after mining
- Mining via `docker exec` into the running network launcher container (no port mapping needed)

## Test plan
- [ ] `motoko-basic_bitcoin` CI passes (ubuntu-24.04, amd64)


raymondk (approved):

motoko/basic_bitcoin/backend/app.mo
```diff
@@ -0,0 +1,134 @@
...
+        tip_height : Nat32;
+        block_headers : [Blob];
+      };
+    } = actor ("aaaaa-aa");
+    await management_actor.bitcoin_get_block_headers({
+      network = NETWORK;
+      start_height;
+      end_height;
+    });
+  };
```
gregorydemay, line 84: 🤖 This `get_block_headers` reimplements `BitcoinApi.get_block_headers` (BitcoinApi.mo:92) inline against the **management canister** `aaaaa-aa`, which leaves the `BitcoinApi` version as dead code (never called) and diverges from this PR's own approach of calling the Bitcoin canister directly at its network-specific principal. Three concerns:

1. **Dead code**: `BitcoinApi.get_block_headers` is now unreachable.
2. **Inconsistency**: every other Bitcoin endpoint goes direct to `g4xu7-…`/`ghsi2-…`; only this one reverts to the management proxy.
3. **Likely missing cycles**: `bitcoin_get_block_headers` is a paid endpoint. Every sibling call attaches cycles (e.g. `100_000_000`); this inline call attaches none, so it may trap with an out-of-cycles error. There's no `get_block_headers` test in the Makefile, so CI won't catch it.

The justifying comment doesn't hold up: `BitcoinApi.get_block_headers` already takes `Types.Network` (which includes `#regtest`), not `mo:ic`'s `BitcoinNetwork`, so the `#regtest` type-error rationale doesn't apply. Suggest deleting the inline actor and calling `BitcoinApi.get_block_headers(NETWORK, start_height, end_height)` to match the other endpoints. If it must stay inline, attach cycles and document why.

motoko/basic_bitcoin/backend/app.mo
```diff
@@ -0,0 +1,134 @@
...
+import P2tr "P2tr";
+import Types "Types";
+import Utils "Utils";
+
+actor class BasicBitcoin(network : Types.Network) {
+
+  /// The Bitcoin network to connect to.
+  ///
+  /// When developing locally this should be `regtest`.
+  /// When deploying to the IC this should be `testnet`.
```
gregorydemay: 🤖 Stale comment: the canister now supports mainnet/testnet/regtest explicitly via `init_args` per environment. "When deploying to the IC this should be `testnet`" predates the three-environment setup and is misleading now (production deploys mainnet).

motoko/basic_bitcoin/backend/app.mo
```diff
@@ -0,0 +1,134 @@
...
+  public func get_balance(address : Types.BitcoinAddress) : async Types.Satoshi {
+    await BitcoinApi.get_balance(NETWORK, address);
+  };
+
+  /// Returns the UTXOs of the given Bitcoin address.
+  public func get_utxos(address : Types.BitcoinAddress) : async Types.GetUtxosResponse {
+    await BitcoinApi.get_utxos(NETWORK, address);
+  };
+
+  /// Returns the 100 fee percentiles measured in millisatoshi/vbyte.
```
gregorydemay: 🤖 Unit mismatch: this doc says `millisatoshi/vbyte` but the type was renamed to `MillisatoshiPerByte`. Pick one unit consistently across the doc comment and the type name.

motoko/basic_bitcoin/docker/start.sh
```diff
@@ -0,0 +1,30 @@
...
+  -rpcuser=ic-btc-integration -rpcpassword=ic-btc-integration \
+  getblockcount >/dev/null 2>&1; do
+  sleep 0.5
+done
+
+echo "bitcoind ready on regtest"
+
+# Hand off to the IC network launcher.
+# --bitcoind-addr wires the IC Bitcoin subnet to our local bitcoind.
+# Port 18443 is the RPC port (used in Makefile via curl JSON-RPC).
```
gregorydemay: 🤖 Misleading comment: this says port 18443 RPC is "used in Makefile via curl JSON-RPC", but the Makefile mines via `docker exec … bitcoin-cli`, not curl JSON-RPC over 18443.

gregorydemay (approved):
Minor comments from Claude, but otherwise Claude is happy with it 😄

motoko/basic_bitcoin/backend/app.mo
```diff
@@ -0,0 +1,116 @@
...
+import Blob "mo:core/Blob";
+
+import BitcoinApi "BitcoinApi";
+import P2pkh "P2pkh";
+import P2trKeyOnly "P2trKeyOnly";
+import P2tr "P2tr";
+import Types "Types";
+import Utils "Utils";
+
+actor class BasicBitcoin(network : Types.Network) {
```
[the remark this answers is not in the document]
marc0olo: Not applicable. `BasicBitcoin` is the main canister entry point — the `actor class` syntax is used here solely to accept the `network` init arg, not to support multiple instantiations. `--default-persistent-actors` applies to it exactly as it would to a plain `actor BasicBitcoin { ... }`.

The AGENTS.md caveat about explicit `persistent` only applies to actor class **sub-WASMs** — classes that are dynamically instantiated to create child canisters (e.g. `Bucket` in composite_query, which gets compiled into a separate WASM per instantiation). `BasicBitcoin` is compiled as a single top-level WASM; there is no sub-WASM here.

The `transient` declarations are also intentional: `KEY_NAME` and `DERIVATION_PATH` are derived values recomputed from the stable `NETWORK` field on each upgrade rather than stored redundantly.

PR #1389, opened 2026-06-12
chore: migrate motoko/icp_transfer to icp-cli

marc0olo:
## Summary

Migrates `motoko/icp_transfer` from dfx to icp-cli, refocused to highlight the AccountIdentifier format and its relationship to principals — the key concept that distinguishes ICP transfers from ICRC-1 transfers and the reason centralized exchanges still require AccountIdentifier.

### Changes

- **`icp.yaml`**: replaces `dfx.json`; uses `@dfinity/motoko@v5.0.0` recipe
- **`backend/app.mo`**: three public functions:
  - `toAccountIdHex(principal, subaccount)` — query returning the AccountIdentifier as a 64-char hex string (block explorer / CEX deposit format); uses the `hex` mops library
  - `transferToPrincipal(amount, principal, subaccount)` — calls `toLedgerAccount` internally; use when you have a principal
  - `transferToAccountId(amount, accountId)` — accepts the raw blob directly; use when an exchange gives you the destination as an AccountIdentifier
- **`mops.toml`**: moc 1.9.0, core 2.5.0, hex 1.0.3, `--default-persistent-actors`
- **`Makefile`**: 3 tests, all balance checks via `icp token balance -q --of-principal` + `awk` for e8s conversion:
  - Test 1: asserts `icp identity account-id --format ledger` equals `toAccountIdHex` — CLI and backend agree
  - Test 2: `transferToPrincipal` — 99_990_000 e8s amount + 10_000 e8s fee = exactly 1 ICP deducted; recipient and backend deltas verified
  - Test 3: `transferToAccountId` with the AccountIdentifier as `vec { N : nat8; ... }` — same deduction, proving both transfer paths reach the same account and the 2 ICP funding drains to zero
- **`README.md`**: explains AccountIdentifier, CEX vs wallet format, `created_at_time = null` disables deduplication
- **CI workflow**: `icp_transfer.yml` using `icp-dev-env-motoko:1.0.0`
- **Deleted**: `dfx.json`, `demo.sh`, `src/`, old workflow `motoko-icp-transfer-example.yml`

### Note on vec nat8 in Test 3

Passing `blob "\xx..."` literals to `icp canister call` when the AccountIdentifier hash contains non-ASCII bytes (>0x7F) causes the ICP ledger to fail with a binary Candid parse error. This reproduces on both icp-cli 0.3.2 and 1.0.0, pointing to a bug in the underlying `@dfinity/candid` text parser — `\xx` escapes in blob literals appear to be decoded as Unicode code points rather than raw hex bytes, producing multi-byte UTF-8 sequences for non-ASCII values and inflating the blob beyond 32 bytes.

The `vec { N : nat8; }` decimal notation is unambiguous and avoids this entirely. The conversion from hex is done in pure shell using `fold -w2` and bash hex arithmetic (`$((16#xx))`).

## Test plan

- [x] `icp network start -d && icp deploy` succeeds locally
- [x] `make test` passes all 3 tests locally
- [ ] CI workflow passes (`motoko-icp_transfer` job)


mbjorkqvist (approved):

raymondk (approved):

lwshang:
Quick clarification on the "Note on vec nat8" — I looked into this and the root cause isn't the Candid text parser. It's a shell `echo` portability issue in the round-trip:

```sh
account_id_blob=$(echo "$account_id" | grep -oE 'blob "[^"]*"')
```

`icp canister call` prints the blob as `blob "\2d\0e…\ff…"`. Under POSIX `/bin/sh` (dash, used in CI) and zsh, `echo` interprets the `\HH` sequences as escape codes and mangles the bytes *before* the second call ever parses them. bash (the macOS `/bin/sh`) leaves them intact — which is why it can look platform/version-dependent. Swapping `echo` for `printf '%s'` preserves the bytes.

Why it only broke with non-ASCII bytes: candid's pretty-printer renders an all-printable-ASCII blob as literal characters (no backslashes), but switches to all-`\HH` escapes the moment any byte is >0x7F — that's when there are backslashes for `echo` to corrupt.

Verified with local experiments: the Rust `candid_parser` used by `icp canister call` (same 0.3.1 in both icp-cli 0.3.2 and 1.0.0) parses `blob "\HH…"` to the correct 32 bytes, and a decode → pretty-print → re-parse round-trip is lossless. The corruption happens purely in `echo`; `printf '%s'` round-trips correctly in dash/zsh/bash.

So there's no bug in icp-cli or candid here (and `@dfinity/candid`, the JS library, isn't involved — `icp canister call` uses the Rust crates). The `vec { N : nat8 }` approach is perfectly fine to keep; just flagging that attributing it to the Candid parser is inaccurate.

PR #1393, opened 2026-06-16
chore: migrate rust/low_wasm_memory to icp-cli

marc0olo:
## Summary

Migrates `rust/low_wasm_memory` from dfx to icp-cli, aligned with `motoko/low_wasm_memory`. Also converts the Motoko example's Makefile to a bash script for consistency.

**Removed:** `dfx.json`, both `Makefile` files, old `rust-low_wasm_memory-example.yml`

**Changed:**
- `icp.yaml` (Rust): replaces `dfx.json`; uses `@dfinity/rust@v3.3.0`
- `backend/lib.rs`: source moved from `src/low_wasm_memory_hook/`; package renamed to `backend`; attribute macros updated to modern `ic_cdk::` proc-macro style
- `rust-toolchain.toml`: added with `wasm32-unknown-unknown` target
- Memory limits (`wasm_memory_limit: 8mib`, `wasm_memory_threshold: 2mib`) moved from test scripts into `icp.yaml` settings for both Rust and Motoko — applied at deploy time, no separate configure step needed
- `rust/low_wasm_memory/test.sh` (replaces `Makefile`): polls `get_executed_functions_order` up to 60s for `OnLowWasmMemory`
- `motoko/low_wasm_memory/test.sh` (replaces `Makefile`): same structure, Motoko naming (`getExecutedFunctionsOrder` / `onLowWasmMemory`)
- `README.md` (both): updated to `bash test.sh`; documents the configured limits
- CI: `low_wasm_memory.yml` — both jobs use `bash test.sh`; images `icp-dev-env-rust:1.0.1` and `icp-dev-env-motoko:1.0.1`

**Alignment with motoko/low_wasm_memory:** Both examples use identical memory limits and the same polling test structure. Naming differences are idiomatic per language (snake_case Rust vs camelCase Motoko).

## Test plan

- [ ] `icp network start -d && icp deploy && bash test.sh` passes locally (Rust)
- [ ] `icp network start -d && icp deploy && bash test.sh` passes locally (Motoko)
- [ ] CI jobs `rust-low_wasm_memory` and `motoko-low_wasm_memory` pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1394, opened 2026-06-16
chore: migrate rust/receiving-icp to icp-cli

marc0olo:
## Summary

Migrates `rust/receiving-icp` to icp-cli and showcases per-environment canister configuration via the `icp.yaml` `environments` block.

### Changes

- **`icp.yaml`**: replaces `dfx.json`; uses `@dfinity/rust@v3.3.0`. Defines three environments with per-canister `environment_variables`:
  - `local` → `ICP_LEDGER_CANISTER_ID = ryjl3-tyaaa-aaaaa-aaaba-cai` (pre-deployed by icp-cli)
  - `staging` → `ICP_LEDGER_CANISTER_ID = xafvr-biaaa-aaaai-aql5q-cai` (TESTICP ledger)
  - `production` → `ICP_LEDGER_CANISTER_ID = ryjl3-tyaaa-aaaaa-aaaba-cai` (mainnet ICP ledger)
- **`backend/lib.rs`**: replaces hardcoded TESTICP principal with `ic_cdk::api::env_var_value("ICP_LEDGER_CANISTER_ID")` — reads the canister environment variable at **runtime** via the system API (the variable is a canister setting applied at deploy time, same mechanism as `Runtime.envVar` in Motoko); `ic_cdk::export_candid!()` added for Candid interface extraction
- **`Cargo.toml`**: ic-cdk 0.18 → 0.20, ic-ledger-types 0.15 → 0.16
- **`test.sh`** (replaces `Makefile`): bash script; 7 tests plus idempotent delta-based balance checks
- **`README.md`**: environment table, deploy commands per environment
- **CI**: `receiving-icp.yml` (renamed from `receiving_icp.yml` to match example dir convention) using `ghcr.io/dfinity/icp-dev-env-rust:1.0.1`
- **Deleted**: `dfx.json`, old workflow

### Key concept: environment-specific canister configuration

Deploying with `icp deploy --environment staging` targets the TESTICP ledger without any code change — icp-cli reads the `environments` block in `icp.yaml` and applies `ICP_LEDGER_CANISTER_ID` as a canister setting. The canister reads it at runtime via `ic_cdk::api::env_var_value`. This pattern applies to any per-environment constant (ledger principals, feature flags, external canister IDs).

## Test plan

- [x] `icp network start -d && icp deploy && bash test.sh` passes locally
- [ ] CI job `rust-receiving_icp` passes


.github/workflows/receiving-icp.yml
```diff
@@ -0,0 +1,28 @@
...
+    env:
+      ICP_CLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - name: Deploy and test
+        working-directory: rust/receiving-icp
+        run: |
+          icp network start -d
+          icp deploy
+          bash test.sh
```
mbjorkqvist, line 28: `AGENTS.md` states *"Every example must have a `Makefile` with a `test` target"*, and the CI template runs `make test`. The precedent for shell-based tests is `rust/threshold-ecdsa`, which keeps a thin Makefile wrapper:

```makefile
test:
	bash test.sh
```

Git Bash compatibility doesn't require dropping the Makefile. To stay consistent with the other examples, you could add a `Makefile` with a `test` target that calls `bash test.sh`, and revert this CI step to `make test`. Or if the idea is to remove `Makefile`s completely, either adjust `AGENTS.md` here, or in a follow-up PR that covers all the examples.

rust/receiving-icp/icp.yaml
```diff
@@ -0,0 +1,30 @@
+canisters:
+  - name: backend
+    recipe:
+      type: "@dfinity/rust@v3.3.0"
```
mbjorkqvist, line 4: Likely missing `configuration: package: backend`. The comparable backend-only example `rust/backend_only` specifies it explicitly:

```yaml
recipe:
  type: "@dfinity/rust@v3.2.0"
  configuration:
    package: backend
```

It may work if the recipe defaults the package name to the canister name, but please confirm this is intentional / that inference works here.

rust/receiving-icp/icp.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# ICP_LEDGER_CANISTER_ID is injected at deploy time by the recipe and baked
```
mbjorkqvist, line 6: Documentation inaccuracy worth fixing since demonstrating this mechanism is the PR's purpose: `ic_cdk::api::env_var_value` reads a **canister environment variable** — a canister *setting* applied at deploy time (the `environment_variables` block under `settings:`) — not WASM metadata. "baked into the WASM" / "embedded as WASM metadata" (also in the PR description) is misleading. Suggest describing it as a canister environment variable read at runtime via the system API.

.github/workflows/receiving-icp.yml
```diff
@@ -0,0 +1,28 @@
+name: receiving_icp
```
mbjorkqvist, line 1: Nit: the convention is `<examplename>.yml`. The example dir is `receiving-icp` (hyphen), so `receiving-icp.yml` would match; this file uses an underscore. Cosmetic only — the internal path filters are correct.

mbjorkqvist (approved):
Thanks @marc0olo! Just a few minor things to point out.

.github/workflows/receiving-icp.yml
marc0olo: there is actually an effort to replace all Makefiles in another PR, see https://github.com/dfinity/examples/pull/1423

that decision was taken after a comment of Michael here: https://github.com/dfinity/examples/pull/1393#discussion_r3428970963

rust/receiving-icp/icp.yaml
marc0olo: this is intentional. we introduced a convention where `package` configuration is optional and only required if the Cargo package name differs from the canister name

rust/receiving-icp/icp.yaml
marc0olo: Fixed — updated lib.rs and README to say 'canister environment variable set at deploy time' rather than 'WASM metadata'. Also updated the PR description.

.github/workflows/receiving-icp.yml
marc0olo: Fixed — renamed to `receiving-icp.yml` to match the example directory name.

lwshang (approved):
Was the removal of the Candid file intentional?

marc0olo:
> Was the removal of the Candid file intentional?

so far yes because it is backend only. but I also thought about it lately again. in any case the candid could be obtained from the canister directly, so I am not sure if we need to include it in the examples if it isn't required for bindings.

PR #1408, opened 2026-06-16
chore: migrate rust/candid_type_generation to icp-cli

marc0olo:
## Summary

- Replaces `dfx.json` with `icp.yaml` using the `@dfinity/rust@v3.3.0` recipe
- Restructures `src/` → `backend/` as a Cargo workspace
- Upgrades to ic-cdk 0.20 (fixes removed `ic_cdk::api::call` API)
- Upgrades ic-cdk-bindgen 0.1.3 → 0.2.0, which generates `Call::bounded_wait` natively (no post-processing needed)
- Rewrites `build.rs` for the 0.2.0 API: `Config::new(name, path).static_callee(principal).set_type_selector_config(toml).generate()`; adds `cargo:rerun-if-changed` directives for fast incremental rebuilds
- Adds `candid/nns_governance.toml` type selector config to preserve `Debug + Clone + serde::Serialize` derives on generated types
- Switches to `include!(OUT_DIR)` pattern in `declarations/mod.rs` — generated file no longer committed to git
- Adds `nns: true` to the local network config in `icp.yaml` so NNS canisters are available at their mainnet IDs locally; the inter-canister call to NNS Governance succeeds in local development and CI
- Adds `test.sh`; test asserts a real successful response (`NNS Governance Neurons`)
- Rewrites `README.md` for the current workflow; explains the full `.did → build.rs → OUT_DIR → include! → use` flow, the type selector config, and the `static_callee` vs `dynamic_callee` choice
- Deletes legacy `rust-candid-type-generation-example.yml` workflow; adds `candid_type_generation.yml` using `ghcr.io/dfinity/icp-dev-env-rust:1.0.1`

## Test plan

- [x] CI job `rust-candid_type_generation` passes
- [x] `icp network start -d && icp deploy && bash test.sh` succeeds locally (returns `NNS Governance Neurons: {...}`)


lwshang (approved):

rust/candid_type_generation/backend/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Inter-canister calls are only allowed from update calls on the IC.
+#[ic_cdk::update]
+async fn list_neurons_pretty() -> String {
+    match fetch_neurons().await {
+        Ok(response) => match serde_json::to_string_pretty(&response) {
+            Ok(json) => format!("NNS Governance Neurons:\n{}", json),
+            Err(_) => format!("NNS Governance Neurons (Debug):\n{:#?}", response),
+        },
+        Err(err) => format!("Error fetching neurons: {}", err),
+    }
```
daniel-wong-dfinity-org-twin: Do as you see fit (including nothing).


```suggestion
    // Fetch neurons from the NNS Governance canister.
    let response = match fetch_neurons().await {
        Ok(response) => ok,
        Err(err) => {
            return format!("Error fetching neurons: {}", err);
        }
    };
    
    // Convert to (pretty) JSON.
    let json = match serde_json::to_string_pretty(&response) {
        Ok(ok) => ok,
        Err(_) => {
            return format!("NNS Governance Neurons (Debug):\n{:#?}", response);
        }
    };
    
    // Generate response (happy case).
    format!("NNS Governance Neurons:\n{}", json)
```

> Flat is better than nested.
> --Zen of Python

Why: using my suggested way, you do not keep indenting every time you add a step to the procedure. Also, it keeps the Err handling code close to where the Err originates.

For an even longer explanation, check out this post in my blog: https://www.tumblr.com/magic-bad/764948183732387840/nesting-bad?source=share

rust/candid_type_generation/backend/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/// Makes a typed inter-canister call to the NNS Governance canister.
+///
+/// `declarations::nns_governance::list_neurons` is a generated function that:
+///   1. Encodes `arg` as Candid
+///   2. Calls the `list_neurons` method on the NNS Governance canister
+///   3. Decodes the response as `ListNeuronsResponse`
```
daniel-wong-dfinity-org-twin: Do as you see fit (including nothing).

Move this content to be directly above the line where list_neurons is called (line 42).

Why: This does not seem like the type of content that belongs in a function's triple slash comments, because it explains not what the function does. Rather, this explains what one of the callees does, how it is generated from a .did file.

rust/candid_type_generation/backend/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+        include_empty_neurons_readable_by_caller: Some(false),
+        include_public_neurons_in_full_neurons: Some(true),
+        page_number: None,
+        page_size: Some(10),
+        neuron_subaccounts: None,
+    };
+
+    declarations::nns_governance::list_neurons(&request)
+        .await
+        .map_err(|e| format!("Inter-canister call failed: {}", e))
```
daniel-wong-dfinity-org-twin: Do as you see fit (including nothing).

I'd delete this, because it obfuscates the return type of list_neurons, and our goal here is to illustrate list_neurons.

rust/candid_type_generation/src/declarations/nns_governance.rs
daniel-wong-dfinity-org-twin, line 1: 😎

rust/candid_type_generation/.gitignore
```diff
@@ -0,0 +1,2 @@
+# ic-cdk-bindgen writes generated Rust types to Cargo's $OUT_DIR at build time.
```
daniel-wong-dfinity-org-twin: > Cargo's $OUT_DIR

And is this the OUT_DIR?

rust/candid_type_generation/.gitignore
```diff
@@ -0,0 +1,2 @@
...
+# Only candid/nns_governance.did is committed; the generated .rs file is not.
```
daniel-wong-dfinity-org-twin: Ok. And how does this fact relate to this dir?

rust/candid_type_generation/backend/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "backend"
+version = "0.1.0"
+edition = "2021"
```
daniel-wong-dfinity-org-twin: I think there is a newer one?

rust/candid_type_generation/backend/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[lib]
+crate-type = ["cdylib"]
```
daniel-wong-dfinity-org-twin, line 8: For a random intelligent Rust programmer (the target audience), this is likely an unfamiliar incantation. Therefore, it probably makes sense to explain that this is how cargo knows that what we want is a .wasm suitable for installation into an ICP canister.

rust/candid_type_generation/backend/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+ic-cdk = "0.20"
+candid = "0.10"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+serde_bytes = "0.11"
```
daniel-wong-dfinity-org-twin: sort

rust/candid_type_generation/backend/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[build-dependencies]
+ic-cdk-bindgen = "0.2.0"
+candid = "0.10"
```
daniel-wong-dfinity-org-twin: Ditto.

rust/candid_type_generation/backend/build.rs
daniel-wong-dfinity-org-twin, line 1: TANGENT: This feels very boilerplate-y. And there is a large amount of it!

But don't do anything about it, because that's out of scope for this PR. This should be fixed by the library...

rust/candid_type_generation/README.md
```diff
@@ -1,103 +1,142 @@
...
+## Fetching and updating Candid definitions
 
-The project uses `ic-cdk-bindgen` as a build dependency to handle the type generation:
+`candid/nns_governance.did` is already checked in and ready to use. To refresh it from the live canister:
 
-```toml
-[build-dependencies]
-ic-cdk-bindgen = "0.1.3"
+```bash
+bash scripts/fetch_candid.sh
```
daniel-wong-dfinity-org-twin, line 91: Is `bash` really necessary here? Why not rely on shebang?

rust/candid_type_generation/README.md
```diff
@@ -1,103 +1,142 @@
...
+```
+
+This runs:
```
daniel-wong-dfinity-org-twin: It's not so helpful to simply copy & paste code from fetch_candid.sh. Instead, it is better to (first) explain what the code DOES. E.g. here, we can say something approximately like, "This updates candid/nns_governance.did.".

The exact code that does this is not important; it is just an implementation detail.

The only other possibly interesting fact is where it sources the new .did file: the canister's "candid:service" metadata, but this is secondary, not as interesting.

rust/candid_type_generation/README.md
```diff
@@ -1,103 +1,142 @@
...
+  > candid/nns_governance.did
 ```
 
-This tool automatically generates idiomatic Rust types with appropriate derives and serialization support.
+After refreshing, rebuild to regenerate the Rust types:
+```bash
+icp build backend
+```
+
+If the Candid interface changed in a breaking way (a method removed, a field type changed), you will get a **compile error** — not a runtime failure. This is the key safety property of the approach: interface mismatches are caught before the canister is ever deployed.
```
daniel-wong-dfinity-org-twin: Ok, but adding an `opt` field is considered a "compatible" Candid change, yet `cargo build` here would most likely break as a result of that change. This is because the generated Rust will have a new field, which would application code that looks like

```
let list_widgets_request = ListWidgetRequest {
    limit: 100,
};
```

Also, you can get problems at run time if `list_widgets` used to return results in chronological order (and you relied on this fact), but then it changed to return results in reverse chronological order. This could happen with zero .did changes. (This ofc is not a realistic change, but I mention it just to illustrate that things can break at run time, even if the .did file itself has NO changes at all.)

rust/candid_type_generation/README.md
```diff
@@ -1,103 +1,142 @@
...
+
+**Reproducibility note**: Always commit the `.did` file. The generated Rust (`$OUT_DIR/nns_governance.rs`) is deterministically derived from it at build time, so committing the `.did` is sufficient to reproduce any historical build exactly. Regenerating from mainnet at an arbitrary time would produce different types if the canister interface had changed, making it impossible to reproduce historical WASM bytes — which matters for any third-party verification.
```
daniel-wong-dfinity-org-twin, line 103: Well, another way is to source .did files from their repo + commit ID (this can be done using a git command). Then, you don't even need your own copy of the .did file, and you can always trace it back to the authoritative source.

rust/candid_type_generation/README.md
```diff
@@ -1,103 +1,142 @@
...
+**Possible enhancement**: Run a bot to fetch the latest `.did` file from mainnet on a regular cadence and open a PR if the project still compiles. This keeps types current with zero manual effort.
+
+## Advantages over manual type copying
+
+|  | Manual copying | `ic-cdk-bindgen` |
+|---|---|---|
+| Keeping types current | Track upstream changes manually | Re-run `fetch_candid.sh` |
+| Missing dependent types | Easy to overlook | Full interface generated |
+| Interface mismatches | Runtime errors | Build-time compile errors |
+| Maintenance | Ongoing manual work | Minimal |
```
daniel-wong-dfinity-org-twin, line 114: Do as you see fit (including nothing).

To me, there is nothing natural about manually transcribing from .did to .rs files. It is a VIOLENT violation of DRY. Therefore, at least to me, this section is preaching to the choir, goes without saying.

OTOH, it isn't really harmful either, so no need to delete if you like it.

rust/candid_type_generation/backend/src/lib.rs
marc0olo: Done — implemented your suggestion. Inlined `fetch_neurons()` into `list_neurons_pretty()`, used early returns for the error paths, and placed the comment explaining the generated function directly at the call site.

rust/candid_type_generation/backend/src/lib.rs
marc0olo: Done — the comment now lives directly above the `list_neurons` call, not in the function's doc comment.

rust/candid_type_generation/backend/src/lib.rs
marc0olo: Done — by inlining the call, the return type of `list_neurons` is now directly visible in the match arm without any wrapping.

rust/candid_type_generation/.gitignore
marc0olo: No — `$OUT_DIR` is a path deep inside Cargo's `target/` directory, outside the project tree entirely. The comment was misleading. Updated it to clarify that generated code goes to `target/` (already covered by the root `.gitignore`), so there is nothing in this directory that needs to be ignored.

rust/candid_type_generation/.gitignore
marc0olo: Same as above — the file now explains that the generated `.rs` ends up in `target/`, not in the project source tree, so this `.gitignore` has no active rules to add. The comment is just explanatory context.

rust/candid_type_generation/backend/Cargo.toml
marc0olo: Done — added a comment explaining that `cdylib` is required to produce a `.wasm` binary for ICP canister deployment.

rust/candid_type_generation/backend/Cargo.toml
marc0olo: Done — sorted both `[dependencies]` and `[build-dependencies]` alphabetically.

rust/candid_type_generation/backend/Cargo.toml
marc0olo: Done.

rust/candid_type_generation/backend/build.rs
marc0olo: Agreed — the boilerplate is a library concern. Leaving it as-is per your suggestion; the `ic-cdk-bindgen` ergonomics can improve independently.

rust/candid_type_generation/README.md
marc0olo: `bash test.sh` is our convention across all migrated examples — it's explicit about the interpreter and consistent. The shebang makes it work either way.

rust/candid_type_generation/README.md
marc0olo: Done — replaced the copy-pasted shell command with a description of what the script does: it fetches the `candid:service` metadata from the NNS Governance canister and writes it to `candid/nns_governance.did`.

rust/candid_type_generation/README.md
marc0olo: Good catch — softened the claim. The updated text now notes that Candid-compatible changes (like adding an optional field) can also require code updates, even though the `.did` itself remains valid.

rust/candid_type_generation/README.md
marc0olo: Good alternative worth noting. For this example we kept the committed `.did` file for simplicity and reproducibility — pinning by git commit would also work and avoids storing the file locally, at the cost of needing network access to fetch at build time.

rust/candid_type_generation/README.md
marc0olo: Keeping it — this is an educational example and the motivation section is useful for developers who haven't encountered bindgen before. Agreed it's obvious once you've seen it, but for someone new to ICP it helps frame why the example exists.

rust/candid_type_generation/backend/Cargo.toml
marc0olo: You were right — updated to `edition = "2024"`. ic-cdk 0.20 itself already uses edition 2024, and `rust-toolchain.toml` doesn't pin a Rust version so the latest stable is used (2024 edition available since Rust 1.85). My earlier reply was wrong.

rust/candid_type_generation/backend/src/lib.rs
```diff
@@ -2,33 +2,14 @@
...
-/// Makes a typed inter-canister call to the NNS Governance canister.
-///
-/// `declarations::nns_governance::list_neurons` is a generated function that:
-///   1. Encodes `arg` as Candid
-///   2. Calls the `list_neurons` method on the NNS Governance canister
-///   3. Decodes the response as `ListNeuronsResponse`
-///
-/// The canister ID (`rrkah-fqaaa-aaaaa-aaaaq-cai`) is embedded as a constant in the
-/// generated code — set in build.rs via `.static_callee(principal)`.
-async fn fetch_neurons() -> Result<ListNeuronsResponse, String> {
```
daniel-wong-dfinity-org-twin, line 31: 😎

rust/candid_type_generation/.gitignore
daniel-wong-dfinity-org-twin: > nothing in this directory that needs to be ignored.

Oh. Then, why not just delete this file entirely? Just so we can explain where ic-cdk-bindgen dumps its output?

daniel-wong-dfinity-org-twin (approved):

PR #1411, opened 2026-06-16
chore: migrate rust/exchange-rates to icp-cli

marc0olo:
## Summary

- Replaces `dfx.json` with `icp.yaml`; adds XRC mock canister (`xrc_mock.wasm.gz` from the 2026.06.26 release) for local testing — the mock returns a fixed rate without making HTTPS outcalls
- Two environments: `local` (deploys backend + XRC mock; icp-cli auto-injects `PUBLIC_CANISTER_ID:xrc`) and `ic` (deploys backend only; injects production XRC canister ID `uf6dk-hyaaa-aaaaq-qaaaq-cai` via env var)
- Moves `src/lib.rs` → `backend/lib.rs`; replaces hardcoded XRC canister ID constant with `PUBLIC_CANISTER_ID:xrc` env var (same pattern as `basic_ethereum` / EVM RPC); adds `ic_cdk::export_candid!()`
- Bumps Cargo to edition 2024
- Adds `test.sh` with a test that now verifies a real numeric rate response from the local mock
- Adds `rust-toolchain.toml` pinning the `wasm32-unknown-unknown` target
- Adds `exchange-rates.yml` CI workflow using `icp-dev-env-rust:1.0.1`
- Deletes `dfx.json`, `src/exchange-rates.did`, `src/lib.rs`

## Test plan

- [x] `icp network start -d && icp deploy && bash test.sh` passes locally
- [x] CI job `rust-exchange-rates` passes


rust/exchange-rates/README.md
```diff
@@ -1,5 +1,50 @@
...
+
+### Deploy to mainnet
+
+To query live exchange rates, deploy to the IC mainnet:
+
+```bash
+icp deploy --environment ic
+icp canister call --environment ic backend get_exchange_rate \
+  '(record { symbol = "ICP"; class = variant { Cryptocurrency } }, \
+    record { symbol = "USD"; class = variant { FiatCurrency } })'
```
mbjorkqvist: IIUC, this command fails as written. The `\` after `Cryptocurrency } },` is inside the single-quoted string, so it isn't a shell line-continuation - a literal backslash + newline is passed into the Candid argument and `icp canister call` won't parse it. How about putting the record on one line (like `test.sh` already does), or moving the continuation outside the quotes?

rust/exchange-rates/README.md
```diff
@@ -1,5 +1,50 @@
...
+```
+
+The local deployment includes the XRC mock canister. `icp-cli` automatically injects the mock's canister ID into the backend via the `PUBLIC_CANISTER_ID:xrc` environment variable.
+
+### Deploy to mainnet
+
+To query live exchange rates, deploy to the IC mainnet:
+
+```bash
+icp deploy --environment ic
```
mbjorkqvist: nit: The other migrated example uses the short form `-e ic` (e.g. `basic_ethereum`).

rust/exchange-rates/backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "backend"
+version = "0.1.0"
+edition = "2024"
```
mbjorkqvist, line 4: nit: `AGENTS.md` prescribes `edition = "2021"` and 18 of 20 migrated Rust examples use 2021 (only `candid_type_generation` uses 2024). Nothing here needs edition 2024 - suggest aligning to 2021 unless there's a specific reason.

mbjorkqvist (approved):
Thanks @marc0olo!

`.github/workflows/ninja_pr_checks.yml` still lists `rust/exchange-rates` and runs `dfx start && dfx deploy` on it, which no longer has a `dfx.json`. This is systemic across all merged migrations (`basic_ethereum`, `canister_logs`, `photo_gallery`, … are all still listed there), so it's consistent with current practice, but I'm flagging it in case it makes sense to address, so that the ninja job isn't being left red for migrated examples.

rust/exchange-rates/README.md
marc0olo: Fixed — put the argument on a single line to avoid the quoting issue.

rust/exchange-rates/README.md
marc0olo: Fixed — switched to `-e ic`.

rust/exchange-rates/backend/Cargo.toml
marc0olo: Updated AGENTS.md to prescribe edition 2024 in this same PR. ic-cdk 0.20 itself already uses edition 2024, and `rust-toolchain.toml` doesn't pin a version so examples always build against the latest stable Rust. Keeping 2021 would mean examples deliberately lag behind the ecosystem they're meant to demonstrate.

rust/exchange-rates/backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[lib]
+# cdylib produces a .wasm binary suitable for deployment as an ICP canister.
+crate-type = ["cdylib"]
+path = "lib.rs"
+
+[dependencies]
+candid = "0.10"
+ic-cdk = "0.20"
+ic-xrc-types = "1.2.0"
+serde = "1.0"
```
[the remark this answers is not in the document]
marc0olo: Fixed — removed `serde` from the direct dependencies.

marc0olo:
@mbjorkqvist regarding the ninja pr checks -> they are currently deactivated. we will reconsider these after the migration is finished and after icp ninja fully supports icp-cli (which will be the case soon :))

lwshang (approved):

PR #1412, opened 2026-06-16
chore: migrate rust/icp_transfer to icp-cli

marc0olo:
## Summary

Migrates `rust/icp_transfer` to icp-cli, aligned with the already-merged `motoko/icp_transfer` (#1389). Also patches `motoko/icp_transfer` (security doc links, hex string API for `transferToAccountId`).

### rust/icp_transfer

- **`icp.yaml`**: replaces `dfx.json`; uses `@dfinity/rust@v3.3.0`
- **`backend/src/lib.rs`**: three public functions matching the Motoko example:
  - `to_account_id_hex(principal, subaccount)` — query returning the AccountIdentifier as a 64-char hex string (block explorer / CEX deposit format)
  - `transfer_to_principal(amount, principal, subaccount)` — calls `AccountIdentifier::new` internally
  - `transfer_to_account_id(amount, account_id_hex)` — accepts a 64-char hex string; uses `AccountIdentifier::from_hex` which validates the CRC32 checksum
- **Cargo.toml**: ic-cdk 0.20, ic-ledger-types 0.16 (0.15 + ic-cdk 0.18 had an `ic-cdk-executor` links conflict; 0.16 requires `^0.19` so both use executor v2)
- **`Makefile`**: 3 tests aligned with the Motoko example — no manual ledger deployment (icp-cli pre-deploys it), no python3:
  - Test 1: `icp identity account-id --format ledger` matches `to_account_id_hex`
  - Test 2: `transfer_to_principal` — asserts exact 1 ICP deduction (99_990_000 amount + 10_000 fee)
  - Test 3: `transfer_to_account_id` via hex string — same deduction, backend drains to zero
- **CI**: `icp-dev-env-rust:1.0.1` (1.0.0 bundled candid-extractor 0.1.4 which fails on ic-cdk ≥ 0.19 due to `ic0::cost_call`; fixed in 0.1.6)
- **Deleted**: `dfx.json`, `demo.sh`, `src/`, old workflow

### motoko/icp_transfer (follow-up patch to #1389)

- `transferToAccountId` now accepts a 64-char hex string instead of a raw blob — consistent with Rust, ergonomic for CEX integrations, uses `mo:hex` for decoding; note on CRC32 library gap added
- Security doc links updated to specific anchors (`inter-canister-calls/#securely-handle-traps-in-callbacks`, `data-integrity-and-authenticity/#certified-variables`)

## Test plan

- [x] `icp network start -d && icp deploy && make test` passes locally (rust)
- [x] `icp network start -d && icp deploy && make test` passes locally (motoko)
- [x] Verified locally with `icp-dev-env-rust:1.0.1` Docker image
- [ ] CI workflow passes (`rust-icp_transfer` and `motoko-icp_transfer` jobs)


lwshang (approved):

rust/icp_transfer/Makefile
```diff
@@ -0,0 +1,62 @@
...
+	    "(record { e8s = 99_990_000 : nat64 }, \"$$cli_hex\")"); \
+	  echo "  result: $$result"; \
+	  echo "$$result" | grep -q 'Ok' || (echo "FAIL: transfer rejected" && exit 1); \
+	  backend_after=$$(get_balance "$$backend"); \
+	  recipient_after=$$(get_balance "$$recipient"); \
+	  echo "  backend:   $$backend_before → $$backend_after  (delta $$((backend_after - backend_before)))"; \
+	  echo "  recipient: $$recipient_before → $$recipient_after  (delta $$((recipient_after - recipient_before)))"; \
+	  [ "$$((backend_after - backend_before))" -eq -100000000 ] || (echo "FAIL: expected backend delta -100000000 e8s" && exit 1); \
+	  [ "$$((recipient_after - recipient_before))" -eq 99990000 ] || (echo "FAIL: expected recipient delta +99990000 e8s" && exit 1); \
+	  echo "  PASS — same recipient reached via AccountIdentifier blob as via principal"
```
mbjorkqvist: This seems to be a stale copy-paste from the blob-based version: this test reaches the recipient via a hex string, not a blob. The Motoko counterpart was fixed in this PR  - would it make sense to keep these two consistent?

```
	  echo "  PASS — same recipient reached via AccountIdentifier hex as via principal"
```

rust/icp_transfer/icp.yaml
```diff
@@ -0,0 +1,8 @@
+networks:
```
mbjorkqvist: The merged `motoko/icp_transfer/icp.yaml` has no `networks:` block and its CI passes, so `managed` appears to be the default and this block looks redundant. Would it make sense to either drop it to keep the two siblings aligned, or add it to the Motoko variant if it's actually load-bearing for the Rust recipe?

mbjorkqvist (approved):
Thanks @marc0olo, just a couple of nits!

rust/icp_transfer/Makefile
marc0olo: Fixed — updated to "hex" to match the Motoko version.

rust/icp_transfer/icp.yaml
marc0olo: Dropped — `managed` is the default and the Motoko sibling omits the block. Now consistent.

marc0olo: it actually doesn't hurt to include it. but it's also not worth adding it. I prefer consistency, thanks!

PR #1417, opened 2026-06-16
chore: migrate rust/stake_neuron_from_cli to icp-cli

marc0olo:
## Summary

- Replace `dfx.json` with `icp.yaml` (`nns: true`) — icp-cli is used only for local network management; the Rust binary connects to the IC directly via `ic-agent` and does not depend on icp-cli at runtime
- Upgrade `ic-agent` `0.35` → `0.47` (add `pem` + `ring` features; `ring` is required for secp256k1 keys generated by icp-cli)
- Upgrade `ic-ledger-types` `0.15` → `0.16`; drop `hex` and `crc32fast` (covered by `ic-ledger-types`)
- Update edition to `"2024"`
- Update `src/main.rs`: rename `--amount` → `--amount-e8s`; add `--compute-only` flag that derives and prints the staking subaccount without making any IC calls — lets users verify the destination before committing funds; update default URL to `localhost:8000`
- Update `src/lib.rs`: use `.unwrap()` for constant principal; clarify that the ledger transfer `memo` is cosmetic and does not need to match the governance `memo` — what matters is the nonce used to derive the staking subaccount; document that anyone can complete step 2; remove debug noise; expose `compute_neuron_staking_subaccount` as `pub`
- Update `rust-toolchain.toml`: remove `channel = "1.80"` pin (kept targets only)
- Replace `setup_and_run.sh` with `test.sh`: creates a funded test identity via icp-cli, builds the binary, runs four tests (subaccount determinism, nonce isolation, below-minimum guard, full staking roundtrip)
- Add CI workflow `.github/workflows/stake_neuron_from_cli.yml` (`icp-dev-env-rust:1.0.1`); remove legacy dfx-based workflow
- Rewrite `README.md`: clarify that `src/lib.rs` is the primary artifact (ic-agent Rust code); icp-cli is local-dev only and not needed on mainnet; explain why Node.js is needed with link to [alternative installation methods](https://cli.internetcomputer.org/1.0/guides/installation/#alternative-installation-methods); remove "domain-separated" jargon; correct nonce/memo claim; add nns.internetcomputer.org link; note ICRC-2 as safer alternative with link to dfinity/icp-cli#637
- Delete `dfx.json`, `setup_and_run.sh`, legacy CI workflow

## Test plan

- [ ] `icp network start -d && bash test.sh` passes — subaccount determinism, nonce isolation, below-minimum guard, full 1 ICP staking roundtrip verified against local NNS
- [ ] `cargo build --release` succeeds (no icp-cli needed)
- [ ] CI workflow `stake_neuron_from_cli` passes on push


rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+/// - The minimum staking amount enforced by NNS Governance is 100_000_000 e8s (1 ICP).
+/// - The local network must be started with `nns: true` in `icp.yaml` so that the
+///   ICP Ledger and NNS Governance canisters are available at their mainnet IDs.
+#[update]
+async fn stake_neuron(amount: u64, nonce: u64) -> Result<StakeNeuronResult, String> {
+    let canister_id = ic_cdk::api::canister_self();
+
+    // The canister holds the ICP and becomes the neuron controller, so the
+    // staking subaccount is derived from the canister's own principal.
+    let subaccount = compute_neuron_staking_subaccount(canister_id, nonce);
```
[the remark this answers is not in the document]
marc0olo: Fixed in 3061b70f — added an early guard that returns an error before any transfer is attempted when `amount < 100_000_000` e8s. This prevents ICP from being stranded at the governance staking subaccount.

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+/// - The canister must hold at least `amount + 10_000` e8s of ICP
+///   (10_000 e8s covers the standard transfer fee).
+/// - The minimum staking amount enforced by NNS Governance is 100_000_000 e8s (1 ICP).
+/// - The local network must be started with `nns: true` in `icp.yaml` so that the
+///   ICP Ledger and NNS Governance canisters are available at their mainnet IDs.
+#[update]
+async fn stake_neuron(amount: u64, nonce: u64) -> Result<StakeNeuronResult, String> {
+    let canister_id = ic_cdk::api::canister_self();
+
+    // The canister holds the ICP and becomes the neuron controller, so the
```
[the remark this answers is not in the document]
marc0olo: Acknowledged — added a code comment noting that `stake_neuron` is unguarded and production deployments should restrict it to an authorized principal. For this example canister, adding a full auth system would obscure the core teaching point (the subaccount computation and two-step staking flow). Other examples with unguarded transfer endpoints (`icp_transfer`) follow the same pattern. The security section in the README also covers this.

rust/stake_neuron_from_cli/backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "backend"
+version = "0.1.0"
+edition = "2024"
+
```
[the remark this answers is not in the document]
marc0olo: Intentional — this repo is migrating new Rust examples to edition 2024. Keeping it.

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    // Step 2: Claim the neuron — Governance checks that it received ICP at
+    // hash(controller, memo) and mints the neuron.
+    let claim_request = ClaimOrRefreshNeuronFromAccount {
+        controller: Some(canister_id),
+        memo: nonce,
+    };
+
+    let claim_response: ClaimOrRefreshNeuronFromAccountResponse =
+        Call::bounded_wait(governance_principal, "claim_or_refresh_neuron_from_account")
+            .with_arg(claim_request)
```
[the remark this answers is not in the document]
marc0olo: Accepted in cc128971 — Copilot's concern about compilation was a false alarm (`with_arg` accepts `impl CandidType + Serialize`, satisfied by both owned and borrowed values), but passing by reference is slightly more idiomatic and consistent with the rest of the repo, so we applied the change.

rust/stake_neuron_from_cli/backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[lib]
+crate-type = ["cdylib"]
```
daniel-wong-dfinity-org-twin: Explain?

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
+use candid::{CandidType, Deserialize, Principal};
+use ic_cdk::call::Call;
+use ic_cdk::{query, update};
```
daniel-wong-dfinity-org-twin: consolidate

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#[derive(CandidType, Deserialize, Debug)]
+enum ClaimOrRefreshResult {
+    Error(GovernanceError),
+    NeuronId(NeuronId),
+}
+
+#[derive(CandidType, Deserialize, Debug)]
+struct ClaimOrRefreshNeuronFromAccountResponse {
+    result: Option<ClaimOrRefreshResult>,
+}
```
daniel-wong-dfinity-org-twin: Put these in a separate file? Or generate from .did?

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+/// - The canister must hold at least `amount + 10_000` e8s of ICP
+///   (10_000 e8s covers the standard transfer fee).
+/// - The minimum staking amount enforced by NNS Governance is 100_000_000 e8s (1 ICP).
+/// - The local network must be started with `nns: true` in `icp.yaml` so that the
+///   ICP Ledger and NNS Governance canisters are available at their mainnet IDs.
+// NOTE: stake_neuron is unguarded — any caller can trigger a transfer from the
+// canister's ICP balance. Production deployments should restrict this to an
+// authorized principal (e.g. the canister controller).
+#[update]
+async fn stake_neuron(amount: u64, nonce: u64) -> Result<StakeNeuronResult, String> {
```
daniel-wong-dfinity-org-twin:
```suggestion
async fn stake_neuron(amount_e8s: u64, nonce: u64) -> Result<StakeNeuronResult, String> {
```

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        return Err("amount must be at least 100_000_000 e8s (1 ICP)".to_string());
+    }
+
+    let canister_id = ic_cdk::api::canister_self();
+
+    // The canister holds the ICP and becomes the neuron controller, so the
+    // staking subaccount is derived from the canister's own principal.
+    let subaccount = compute_neuron_staking_subaccount(canister_id, nonce);
+    let governance_principal =
+        Principal::from_text(NNS_GOVERNANCE_CANISTER_ID).map_err(|e| e.to_string())?;
```
daniel-wong-dfinity-org-twin: Do as you see fit.


```suggestion
        Principal::from_text(NNS_GOVERNANCE_CANISTER_ID).unwrap();
```

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    // Step 1: Transfer ICP to the Governance canister's staking subaccount.
+    let transfer_args = ic_ledger_types::TransferArgs {
+        memo: Memo(nonce),
```
daniel-wong-dfinity-org-twin: You might want to say that this field does not need to be set. Otherwise, people will think this is one of the essential pieces of incantation. I mean, such a misconception is not "actually" harmful, but this could be confusing, since the same word ("memo") is used to refer to unrelated things in different contexts.

rust/stake_neuron_from_cli/backend/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    // would reject the claim and strand funds in the staking subaccount.
+    if amount < 100_000_000 {
+        return Err("amount must be at least 100_000_000 e8s (1 ICP)".to_string());
+    }
+
+    let canister_id = ic_cdk::api::canister_self();
+
+    // The canister holds the ICP and becomes the neuron controller, so the
+    // staking subaccount is derived from the canister's own principal.
+    let subaccount = compute_neuron_staking_subaccount(canister_id, nonce);
```
daniel-wong-dfinity-org-twin: Since there are a couple of subaccounts involved here (source vs. destination), it would be a good idea to qualify this name. E.g.


```suggestion
    let new_neuron_subaccount = compute_neuron_staking_subaccount(canister_id, nonce);
```

rust/stake_neuron_from_cli/README.md
```diff
@@ -1,121 +1,85 @@
-# Neuron Staking from CLI
+# Neuron Staking
```
daniel-wong-dfinity-org-twin: The name of the dir still contains "cli". I mean, I'm not sure why this example uses a canister. Seems tangential. I would say the main thing we want to show here is that it's a 2 step dance:

1. Send ICP (to the right place) by calling the Ledger canister.
2. Call Governance canister's `claim_or_refresh` method.

Doing this from a canister adds another layer, which is possibly confusing.

One thing that you might want to point out is that anyone can do step 2 to complete the process, because the result is already fully determined from step 1. The neuron still ends up being controlled by the principal associated with the neuron's subaccount, regardless of who the caller is in step 2.

rust/stake_neuron_from_cli/README.md
daniel-wong-dfinity-org-twin, line 1: You might want to point people to nns.internetcomputer.org. There, they can create neurons, vote on proposals, and earn rewards in a UI. Going through the UI first might help to introduce the subject/concepts. Ofc, this is for people who want to create neurons programmatically (in Rust). I'm just saying, before they become interested in that, they maybe want an interactive introduction.

rust/stake_neuron_from_cli/README.md
```diff
@@ -1,121 +1,85 @@
...
+The neuron staking subaccount is the most critical thing to get right when staking a neuron.
+If you send your ICP to the wrong subaccount, the ICP will be permanently lost, as there will
+not be a way to ask the Governance canister to retrieve it.
 
-Therefore, test your implementation (like in this example project) against a test version of Governance to ensure
-that it is able to send the ICP to the right destination.
+Therefore, test your implementation (like in this example project) against a known reference
+before sending real ICP.
+
+The subaccount is a domain-separated SHA-256 hash over the controller principal and nonce:
```
daniel-wong-dfinity-org-twin: I think many people don't necessary know the term "domain-separated", and it isn't essential here. Therefore, I would either omit it entirely, or mention it much less prominently. The main point here is, here is what you need to feed into SHA-256:

1. 12
2. "neuron-stake"
3. principal ID of the neuron's controller (as a blob, not the text form)
4. nonce aka memo (big-endian bytes)

rust/stake_neuron_from_cli/README.md
```diff
@@ -1,121 +1,85 @@
...
-)
-.with_arg(Encode!(&claim_request)?)
-.call_and_wait()
-.await?;
-```
+1. **Transfer ICP** — Transfer tokens to NNS Governance at the computed subaccount
+2. **Claim Neuron** — Call `claim_or_refresh_neuron_from_account` to register the transfer as a neuron
 
-### 3. **Identity Management with ic-agent**
+The nonce used in step 1 (as the transfer memo) must match the nonce used in step 2.
```
daniel-wong-dfinity-org-twin: I don't think so. It's something that I think nns-dapp does, and can be helpful, but it is not required.

What matters is the memo that you used when deriving the subaccount. You can refer to a neuron by controller + memo. When you do that, Governance re-constructs the subaccount using the same scheme, and Governance can map subaccounts to neurons.

rust/stake_neuron_from_cli/README.md
```diff
@@ -1,121 +1,85 @@
...
-```rust
-async fn load_identity(identity_path: &PathBuf) -> Result<Box<dyn Identity>, Box<dyn std::error::Error>> {
-    let pem_content = std::fs::read_to_string(identity_path)?;
+### Prerequisites
 
-    // Try different identity formats
-    if let Ok(identity) = BasicIdentity::from_pem(pem_content.as_bytes()) {
-        return Ok(Box::new(identity));
-    }
+- [Node.js](https://nodejs.org/) v18+
```
daniel-wong-dfinity-org-twin: Is this really required?

rust/stake_neuron_from_cli/test.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "PASS" || (echo "FAIL: subaccount hash does not match expected reference value" && exit 1)
+
+echo "=== Test 2: compute_subaccount changes with nonce ==="
+s0=$(icp canister call --query backend compute_subaccount "(principal \"$backend\", 0 : nat64)")
+s1=$(icp canister call --query backend compute_subaccount "(principal \"$backend\", 1 : nat64)")
+echo "Nonce 0: $s0"
+echo "Nonce 1: $s1"
+[ "$s0" != "$s1" ] && echo "PASS" || (echo "FAIL: different nonces produced the same subaccount" && exit 1)
+
+echo "=== Test 3: stake_neuron rejects amount below 1 ICP minimum ==="
```
daniel-wong-dfinity-org-twin, line 37: But isn't the bigger problem here that canister has no ICP yet?? I mean, even if the minimum were 1 e8s, this would inevitably end up failing in some OTHER way, simply because the canister doesn't HAVE 99_999_999 e8s (+ transfer fee) to spend on neuron creation.

rust/stake_neuron_from_cli/test.sh
```diff
@@ -0,0 +1,44 @@
...
+backend=$(icp canister status backend -i)
+echo "Backend canister: $backend"
+
+echo "=== Test 1: compute_subaccount matches known reference value ==="
+# Fixed reference: principal "aaaaa-aa" (management canister), nonce 0.
+# The expected value is a pre-computed SHA-256 domain-separated hash.
+# If this changes, the subaccount computation is broken and real ICP could be sent
+# to the wrong address.
+result=$(icp canister call --query backend compute_subaccount \
+  '(principal "aaaaa-aa", 0 : nat64)')
```
daniel-wong-dfinity-org-twin: This is not a very realistic value. E.g. nobody would ever try to create a neuron with "aaaaa-aa" being the controller! I mean, maybe it still gets the point across, but I can see this as being confusing.

daniel-wong-dfinity-org-twin (approved):
We now support ICRC-2, which is much nicer, because if there is a crash between steps 1 and 2, the ICP is still in the user's control. Here is a good starting point: https://sourcegraph.com/r/github.com/dfinity/ic@cc1287f38c519256d7189d0730fb5f2c145fb27a/-/blob/rs/nns/governance/canister/governance.did?L1576-1593

marc0olo:
Hi @daniel-wong-dfinity-org-twin — thanks for the detailed review, it was very helpful. Here's what we acted on:

**Architectural change**: You were right to flag the canister approach. The original example was a CLI binary using `ic-agent` — we had inadvertently converted it to a canister during migration, which changed its fundamental purpose. We've reverted back to the CLI binary (`src/lib.rs` + `src/main.rs`), with `icp.yaml` used only to start a local NNS replica for testing. The example now correctly shows how to stake ICP from off-chain Rust code.

**Code changes applied**:
- Renamed `amount` → `amount_e8s` ✓
- Renamed `subaccount` → `new_neuron_subaccount` ✓
- `.unwrap()` for the constant governance principal ✓
- Clarified the `memo` field: the ledger transfer memo is cosmetic; what matters is that the nonce used to derive the staking subaccount matches the `memo` in `claim_or_refresh_neuron_from_account` ✓
- Added `--compute-only` flag so users can verify the staking subaccount before committing any ICP ✓
- Documented that anyone can complete step 2 ✓

**README updates**:
- Removed "domain-separated" jargon; listed the 4 SHA-256 inputs plainly ✓
- Corrected the nonce/memo claim ✓
- Added nns.internetcomputer.org link ✓
- Made clear that `src/lib.rs` is the primary artifact and icp-cli is local-dev only ✓
- Added ICRC-2 note pointing to https://github.com/dfinity/icp-cli/pull/637 for the safer atomic approach once icp-cli supports it ✓

**Not applied**: governance types in a separate file — for an example this size, keeping everything in one `lib.rs` is more readable.

Would you mind having a quick re-review? We'd love your sign-off before merging.

rust/stake_neuron_from_cli/Cargo.toml
```diff
@@ -1,3 +1,19 @@
...
+[dependencies]
+candid = "0.10"
+# pem: enables from_pem on identity types.
+# ring: enables Secp256k1Identity (needed for keys generated by icp-cli).
+ic-agent = { version = "0.47", features = ["pem", "ring"] }
+ic-ledger-types = "0.16"
+clap = { version = "4", features = ["derive"] }
+sha2 = "0.10"
+serde = { version = "1.0", features = ["derive"] }
+tokio = { version = "1", features = ["full"] }
```
daniel-wong-dfinity-org-twin, line 19: sort

rust/stake_neuron_from_cli/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    let pem = std::fs::read(&args.identity_path)?;
+    // Try Ed25519 first, then secp256k1 (default key type created by icp-cli).
+    let identity: Box<dyn Identity> = if let Ok(id) = BasicIdentity::from_pem(&pem) {
+        Box::new(id)
+    } else {
+        Box::new(Secp256k1Identity::from_pem(&pem)?)
+    };
+    let controller = identity.sender()?;
```
daniel-wong-dfinity-org-twin, line 34: Do as you see fit.

This code is repeated in main.rs (but this is ok, because we are just trying to show people how to do stuff).

rust/stake_neuron_from_cli/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    let agent = Agent::builder()
+        .with_url(&args.ic_url)
+        .with_identity(identity)
+        .build()?;
+    // fetch_root_key is only needed for local networks; on mainnet the root key is
+    // compiled into ic-agent and this call is a no-op.
+    agent.fetch_root_key().await?;
```
daniel-wong-dfinity-org-twin: In the [ic_agent docs], it says

[ic_agent docs]: https://docs.rs/ic-agent/latest/ic_agent/struct.Agent.html

```
 // Only do the following call when not contacting the IC main net (e.g. a local emulator).
 // This is important as the main net public key is static and a rogue network could return
 // a different key.
 // If you know the root key ahead of time, you can use `agent.set_root_key(root_key);`.
 agent.fetch_root_key().await?;
```

Therefore, I think you need to avoid calling fetch_root_key when talking to the real IC.

rust/stake_neuron_from_cli/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    let governance = Principal::from_text(NNS_GOVERNANCE_CANISTER_ID).unwrap();
+    let transfer_args = TransferArgs {
+        // The memo here is only a label on the ledger transaction; it does not need
+        // to match the nonce passed to claim_or_refresh_neuron_from_account below.
+        memo: Memo(args.nonce),
+        amount: Tokens::from_e8s(args.amount_e8s),
+        fee: Tokens::from_e8s(10_000),
+        from_subaccount: None,
+        to: AccountIdentifier::new(&governance, &new_neuron_subaccount),
+        created_at_time: None,
```
daniel-wong-dfinity-org-twin, line 68: Do as you see fit.

Setting this is useful for deduplication. You don't have to show that, because it is more advanced, but maybe add a comment and point to documentation.

rust/stake_neuron_from_cli/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    hasher.update(controller.as_slice());
+    hasher.update(nonce.to_be_bytes());
+    Subaccount(hasher.finalize().into())
+}
+
+fn hex_encode(bytes: &[u8]) -> String {
+    bytes.iter().map(|b| format!("{b:02x}")).collect()
+}
+
+// ── Candid types for NNS Governance ─────────────────────────────────────────
```
daniel-wong-dfinity-org-twin, line 142: Let's point to the governance.did file.

rust/stake_neuron_from_cli/test.sh
daniel-wong-dfinity-org-twin: Looks like this was not addressed? I think if you simply move the `icp token transfer ...` line here (or earlier), this would be addressed.

rust/stake_neuron_from_cli/test.sh
```diff
@@ -1,44 +1,42 @@
...
-echo "$result"
-echo "$result" | grep -q 'Ok' || (echo "FAIL: stake_neuron did not return Ok" && exit 1)
-echo "$result" | grep -q "neuron_id" || (echo "FAIL: missing neuron_id in result" && exit 1)
-echo "$result" | grep -q "$expected_subaccount" && \
-  echo "PASS" || (echo "FAIL: subaccount_hex in result does not match compute_subaccount" && exit 1)
+"$BINARY" --url "$URL" --identity /tmp/stake-neuron-test.pem --amount-e8s 99999999 --nonce 0 2>&1 | \
+  grep -q "100_000_000" && echo "PASS" || (echo "FAIL" && exit 1)
+
+echo "=== Setup: fund test identity with 2 ICP ==="
+icp token transfer 2 "$PRINCIPAL"
```
daniel-wong-dfinity-org-twin: Do as you see fit.

You might want to add a comment that says why 2 instead of 1: we need more than 1 to cover the transfer fee that gets incurred when sending ICP to the Governance canister.

daniel-wong-dfinity-org-twin (approved):

rust/stake_neuron_from_cli/src/lib.rs
marc0olo: yes, I think we should generally avoid calling this function; even locally. locally we can determine the root key via icp-cli and set it 🤔 

in any case, the code should also work locally.

rust/stake_neuron_from_cli/Cargo.toml
marc0olo: Fixed in 758bf93f — dependencies are now sorted alphabetically.

rust/stake_neuron_from_cli/src/lib.rs
marc0olo: Leaving as-is. The identity loading is duplicated in `main.rs` for the `--compute-only` path, but that path intentionally skips building an `Agent` entirely, so it can't reuse the library's `stake_neuron` flow. The duplication is minimal and keeps each path self-contained.

rust/stake_neuron_from_cli/src/lib.rs
marc0olo: Fixed in 758bf93f — replaced the unconditional call with a URL-based guard: `fetch_root_key` is now only called when the URL contains `localhost` or `127.0.0.1`. Updated the comment to explain the security rationale (ic-agent has the real mainnet key compiled in; fetching it dynamically against an untrusted host would allow a rogue node to return a fake key).

rust/stake_neuron_from_cli/src/lib.rs
marc0olo: Fixed in 758bf93f — added a comment pointing to the deduplication docs: https://docs.internetcomputer.org/guides/digital-assets/ledgers/#transaction-deduplication

rust/stake_neuron_from_cli/src/lib.rs
marc0olo: Fixed in 758bf93f — added a link to the full governance.did: https://github.com/dfinity/ic/blob/master/rs/nns/governance/canister/governance.did

rust/stake_neuron_from_cli/test.sh
marc0olo: Fixed in 758bf93f — moved the `icp token transfer 2` funding step to before Test 3 so the identity is funded throughout all tests.

rust/stake_neuron_from_cli/test.sh
marc0olo: Fixed in 758bf93f — added a comment explaining that 2 ICP is needed because the minimum stake is 1 ICP (100_000_000 e8s) and the ledger charges an additional 10_000 e8s transfer fee on top.

raymondk (approved):

PR #1429, opened 2026-07-08
chore: add ic env + build-image.sh to motoko/basic_bitcoin

marc0olo:
## Summary

Migrates `motoko/basic_bitcoin` from dfx environments (`staging`/`production`) to a single `ic` environment, and replaces the `Makefile` with dedicated shell scripts.

### Environment consolidation

- Replaces `staging` + `production` environments in `icp.yaml` with a single `ic` environment using Bitcoin testnet4 and `test_key_1` — the standard default for IC network deployments
- `app.mo` updated to reference `ic` env instead of `staging`/`production`
- README updated to reflect the new environment and document how to switch to mainnet (`key_1`) for production deployments

### Why a single `ic` environment

icp.ninja (and other tools) expect an environment named `ic` when deploying to the IC network. Using `test_key_1` + testnet4 as the default is the safe choice — developers can optionally switch to mainnet by changing the `init_args` in `icp.yaml`, at which point the canister automatically selects `key_1`.

### Tooling

- Replaces the `Makefile` with `build-image.sh` (Docker image build) and `test.sh` (end-to-end tests)
- CI updated to call `bash test.sh` directly

## Test plan

- [ ] `./build-image.sh` builds the Docker image successfully
- [ ] `icp network start -d && icp deploy --cycles 30t && bash test.sh` passes locally
- [ ] `icp deploy -e ic --cycles 30t` deploys to IC against Bitcoin testnet4


motoko/basic_bitcoin/Makefile
```diff
@@ -2,15 +2,7 @@ IMAGE_NAME = icp-cli-network-launcher-bitcoin
 # Find the running container built from our custom image
 BITCOIN_CONTAINER = $(shell docker ps --filter "ancestor=$(IMAGE_NAME)" --format "{{.ID}}" | head -1)
 
-.PHONY: build-image test topup
-
-build-image:
-	# Because we're building off of :latest, use --pull to fetch the latest image.
-	# Remove `--pull` if the Dockerfile is updated to pin the base image version.
-	docker build --pull -t $(IMAGE_NAME) .
```
mbjorkqvist, line 10: nit: The PR description calls `build-image.sh` an "alternative" to `make build-image`, but the target is actually removed here, so it's a replacement, not an alternative.

motoko/basic_bitcoin/build-image.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+# Uses --pull to always fetch the latest base image.
+# Remove --pull if the Dockerfile is updated to pin the base image version.
+docker build --pull -t icp-cli-network-launcher-bitcoin .
```
mbjorkqvist, line 5: nit: The image name `icp-cli-network-launcher-bitcoin` is now hard-coded in two places (here, and as `IMAGE_NAME` in the `Makefile`), so there are two sources of truth to fix if the name ever changes.

mbjorkqvist (approved):
Thanks @marc0olo! Just a couple of nits.

motoko/basic_bitcoin/Makefile
marc0olo: Thanks for the proactive review, Mathias — this PR should have been a draft from the start, sorry about that. You're right on both counts. We've now removed the Makefile entirely: `build-image.sh` and `test.sh` are the full replacements, and the CI calls `bash test.sh` directly.

motoko/basic_bitcoin/build-image.sh
marc0olo: Fixed by removing the Makefile — the image name is now only defined in `build-image.sh`.

raymondk (approved):

PR #1439, opened 2026-07-10
chore: migrate rust/llm_chatbot to icp-cli

marc0olo:
## Summary

Migrates `rust/llm_chatbot` from dfx to icp-cli, and has the backend call the LLM canister's `v1_chat` endpoint directly instead of through a helper crate.

**Tooling / config**
- Replaces `dfx.json` with `icp.yaml`. The `llm` canister is deployed from a pre-built WASM (v0.3.1, Ollama mode); the `ic` environment restricts deployment to `[backend, frontend]` so `icp deploy` does not redeploy the LLM canister on mainnet.
- The backend resolves the LLM canister via the `PUBLIC_CANISTER_ID:llm` env var: `icp deploy` auto-injects it locally (after deploying the pre-built `llm`), and the `ic` environment sets it to the mainnet LLM principal (`w36hm-eqaaa-aaaal-qr76a-cai`) in `icp.yaml`. This matches how other examples inject shared mainnet canisters (e.g. `rust/basic_ethereum` for `evm_rpc`).

**Backend (`backend/`)**
- Calls the LLM canister's `v1_chat` endpoint directly (`lib.rs`), with no external LLM helper crate. The Candid types are inlined, and the LLM canister principal is read from the injected `PUBLIC_CANISTER_ID:llm` env var (see Tooling / config). The optional `tools` field is omitted from the request (Candid decodes an absent optional field as `null`), so no tool types are needed.
- `Cargo.toml`: removes the `ic-llm` dependency, adds `serde`, bumps `ic-cdk` to 0.20, and sets edition 2024.
- Adds `backend.did` (the extracted Candid interface).

**Frontend (`frontend/`)**
- Replaces dfx-generated declarations + `vite-plugin-environment` with `@icp-sdk/bindgen` (Vite plugin) and a new `src/actor.js` that builds the actor from `safeGetCanisterEnv()`.
- Renders assistant responses as Markdown (`react-markdown` + `remark-gfm`, styled with `@tailwindcss/typography`), so bold text, lists, headings, and code blocks display formatted instead of as raw text. User messages stay plain text.
- `package.json`: drops `dfx generate` from `prebuild`; removes `vite-plugin-environment`, `react-json-view-lite`, and the `@types/*` packages; adds `@icp-sdk/bindgen`, `@icp-sdk/core` (6.x), `react-markdown`, `remark-gfm`, and `@tailwindcss/typography`.

**CI**
[...]

PR #1441, opened 2026-07-14
chore: migrate motoko/llm_chatbot to icp-cli

marc0olo:
## Summary

Migrates `motoko/llm_chatbot` from dfx to icp-cli, mirroring the rust variant (#1439) so the two examples stay equivalent apart from the backend language. The backend calls the LLM canister's `v1_chat` endpoint directly instead of through the `mo:llm` helper library.

**Tooling / config**
- Replaces `dfx.json` with `icp.yaml`. The `llm` canister is deployed from a pre-built WASM (v0.3.1, Ollama mode); the `ic` environment restricts deployment to `[backend, frontend]` so `icp deploy` does not redeploy the LLM canister on mainnet.
- `mops.toml`: bumps `moc` to 1.11.0 and `core` to 2.5.0, drops the `mo:llm` dependency, adds `--default-persistent-actors`, and declares the backend canister (`main` + `candid`).

**Backend (`backend/app.mo`)**
- Calls the LLM canister's `v1_chat` endpoint directly using `mo:core`, dropping `mo:llm`. `mo:llm` hardcodes the mainnet canister principal and cannot reach the locally-deployed LLM canister under icp-cli's env-var injection model, so the types are inlined and the principal is resolved from the injected `PUBLIC_CANISTER_ID:llm` env var via `Runtime.envVar<system>` — matching `motoko/evm_block_explorer`. The optional `tools` field is omitted from the request (Candid decodes an absent optional field as `null`), so no tool types are needed.
- Adds `backend/backend.did` (moc-generated). Same public interface as the rust variant.
- Locally, `icp deploy` auto-injects `PUBLIC_CANISTER_ID:llm`; the `ic` environment sets it to the mainnet LLM principal (`w36hm-eqaaa-aaaal-qr76a-cai`) in `icp.yaml`.

**Frontend (`frontend/`)**
- Brought to parity with the migrated rust frontend: `@icp-sdk/bindgen` (Vite plugin) + a new `src/actor.js` using `safeGetCanisterEnv()`, and assistant responses rendered as Markdown (`react-markdown` + `remark-gfm`, styled with `@tailwindcss/typography`) so bold text, lists, headings, and code blocks display formatted instead of as raw text. Latest deps (`@icp-sdk/core` 6.x, `@icp-sdk/bindgen` 0.4.x). The built frontend is byte-identical to the rust variant's.

**CI**
- Adds `.github/workflows/llm_chatbot.yml` with a deploy-only motoko job. As with the rust variant, the backend's `prompt`/`chat` methods require the LLM canister to reach a running Ollama, which CI cannot provide, so there is nothing to assert beyond a successful deploy.

**Removals**
- `dfx.json`, `BUILD.md`, `.devcontainer/`.

> **Note for reviewers:** this workflow file overlaps with #1439, which also adds `.github/workflows/llm_chatbot.yml` (with the rust job). Each PR carries only its own variant's job because the other variant isn't migrated on its branch. Whichever merges second should reconcile them into a single dual-job workflow, like `evm_block_explorer.yml`.

> **Note for reviewers:** end-to-end testing is manual — deploy, then chat in the frontend. This requires a running Ollama (`ollama serve` + `llama3.1:8b`); the README's setup uses `ollama run` to warm the model, since a cold model load can exceed the LLM canister's ~30s outcall deadline on the first call.


marc0olo:
**Re: Copilot review [#4693179623](https://github.com/dfinity/examples/pull/1441#pullrequestreview-4693179623)**

**`test.sh` / CI** — Keeping the workflow deploy-only, intentionally (consistent with the rust variant in #1439). The backend's only methods (`prompt`, `chat`) both require the LLM canister to reach a running Ollama, which CI cannot provide, so there is no public method that can be meaningfully asserted in CI. A test that asserts a failure/reject when Ollama is absent would test a broken path and be brittle against the pre-built LLM canister's reject text and timing. This matches the repo's existing deploy-only examples where CLI assertions aren't feasible (the vetkeys and hosting examples). `icp deploy` still exercises the meaningful regression surface: it compiles the Motoko backend, generates the frontend bindings from `backend.did`, builds the frontend, and deploys the pre-built LLM canister.

raymondk (approved):

PR #1445, opened 2026-07-16
chore(vetkeys): split encrypted_notes_app_vetkd into self-contained motoko + rust examples

marc0olo:
## Summary

Splits the vetkeys **`encrypted_notes_app_vetkd`** example into two self-contained examples (one per language) — same restructure as #1443 / #1444. Previously it lived only under `rust/vetkeys/encrypted_notes_app_vetkd/` with a shared **Svelte** frontend symlinked into `motoko/` and `rust/` subdirs.

```
motoko/vetkeys/encrypted_notes_app_vetkd/   backend/app.mo + own frontend/ + icp.yaml + mops.toml
rust/vetkeys/encrypted_notes_app_vetkd/      backend/ (flattened) + own frontend/ + icp.yaml + Cargo.toml
```

## Frontend (Svelte)
- Duplicated per language; bindings generated by the **`@icp-sdk/bindgen` Vite plugin** from the committed `backend/backend.did` (replaces `scripts/gen_bindings.sh`).
- **Unified the binding imports.** The shared frontend imported the bindings via three inconsistent paths (`declarations/encrypted_notes/encrypted_notes_rust`, `…/backend.did`, and `../lib/backend` — the last of which didn't resolve and only survived as an erased `import type`). All now import from a single `../bindings/backend`.
- Added a **root workspace `package.json`** so `npm run dev` / `npm run build` run from the example root; kept all Svelte deps.

## Backend
- **Motoko:** idiomatic **camelCase** public interface (`createNote`, `getNotes`, `updateNote`, `addUser`, `removeUser`, `deleteNote`, `encryptedSymmetricKeyForNote`, `symmetricKeyVerificationKeyForNote`; `EncryptedNote.encryptedText`). The **vetKD management-canister interface is kept snake_case** (fixed system API). **moc 1.11.0** + `--default-persistent-actors` (dropped `persistent`). Committed `backend/backend.did`.
- **Rust:** unchanged (snake_case); relocated, and the hand-maintained `.did` moved to `backend/backend.did`.

Each frontend matches its own backend (Motoko camelCase / Rust snake_case).

## Docs & CI
- README split per language (cross-linked); prerequisites link each tool with its command inline.
- `test.sh` added (all backend methods here are update calls — deliberately, for certification — so no `--query`).
- `vetkeys-encrypted-notes-app-vetkd.yml`: `motoko` + `rust` jobs on the new paths, watching both, running `test.sh`.

## Note on type-checking
The frontend carries **pre-existing** `svelte-check` type issues (auth-store discriminated-union typing, missing lib type declarations, implicit-any). The build has always been `vite build` (esbuild, no type-check) and never gated on `svelte-check`; this change introduces none of those errors.

## CODEOWNERS

## Verification (local)
- **Motoko:** `vite build` + deploy, `test.sh` PASS, and a created note **survives an upgrade**.

> Part of the vetkeys split series (#1443, #1444). Remaining: `password_manager`, `password_manager_with_metadata`.


marc0olo:
Heads-up on cross-links: the Motoko README's **API level** section now links to the **Motoko** password manager examples (`../password_manager`, `../password_manager_with_metadata`). Those paths only exist once the password-manager split PRs (#1446, #1447) are merged. If this PR merges first, those two links will 404 until #1446/#1447 land — worth ordering the merges accordingly (or I can temporarily point them at the Rust siblings and switch them back afterward).

Also in this update:
- **security-checklist.md** condensed into a focused, language-specific version for each of Motoko and Rust. The previous file was a 165-line identical copy carried over from the older *device-based* encrypted_notes and was inaccurate for the vetKD version — it claimed "no inter-canister calls" (this version calls the vetKD management canister) and referenced `submit_ciphertexts` / device removal that don't exist here. Each README now links its own checklist.
- Dropped "smart contract" web3 jargon from the backend comments ("a single canister smart contract" → "a single canister").

marc0olo:
Thanks @Copilot. I looked into all four points. Three are pre-existing issues carried over verbatim from the original shared frontend (this PR only splits the example into self-contained Motoko/Rust copies — the frontend changes here are limited to import paths, `snake_case`→`camelCase` method names to match the backend, and the `PUBLIC_CANISTER_ID:backend` env key), and the fourth is intentional:

- **`store/notes.ts` — uninitialized `notePollerHandle`**: pre-existing. When it is `undefined` the guard `notePollerHandle !== null` is still true, so `clearInterval(undefined)` runs, which is a no-op in browsers. No functional impact.
- **`store/notes.ts` — unhandled rejection in the poll loop**: pre-existing. It only surfaces as a console warning on an already-failing poll (`showError` on the error path).
- **`lib/actor.ts` — no fail-fast on a missing `canisterId`**: pre-existing. This is a defensive guard rather than a bug; a normal deploy always injects the canister id.
- **CI has no `test.sh`**: intentional. None of these vetKeys examples shipped a `test.sh` on `master`, and CI's default identity is anonymous (`2vxsx-fae`), which several of these backends reject (e.g. `getNotes` asserts a non-anonymous caller) — so a CLI `test.sh` would trap in CI even though it passes locally. The workflow is therefore deploy-only, which still verifies the build and install.

Since none of these were introduced by this PR, I'm leaving the frontend untouched to keep the split faithful and avoid re-testing behavior that already works. They can be addressed separately upstream if desired.

raymondk (approved):

pierugo-dfinity (approved):

PR #1446, opened 2026-07-16
chore(vetkeys): split password_manager into self-contained motoko + rust examples

marc0olo:
## Summary

Splits the vetkeys **`password_manager`** example into two self-contained examples (one per language) — same restructure as #1443–#1445. Previously it lived only under `rust/vetkeys/password_manager/` with a shared **Svelte** frontend symlinked into `motoko/` and `rust/` subdirs.

```
motoko/vetkeys/password_manager/   backend/app.mo + own frontend/ + icp.yaml + mops.toml
rust/vetkeys/password_manager/     backend/ (flattened) + own frontend/ + icp.yaml + Cargo.toml
```

## Why this one differs (no camelCase, no bindgen)

This example uses the **Encrypted Maps** abstraction. The frontend talks to the backend via the **`@icp-sdk/vetkeys` Encrypted Maps client** (`DefaultEncryptedMapsClient`), **not generated bindings** — so there's no icpBindgen step. The backend keeps the Encrypted Maps **library interface (snake_case)**, which the SDK client calls verbatim, so there's **no camelCase conversion** (it's a fixed library API, like the vetKD system API). Both frontends are therefore identical.

## Changes
- **Frontend (Svelte):** per-language copy; canister renamed to **`backend`** (env key `PUBLIC_CANISTER_ID:backend`); dropped eslint/prettier and the `BACKEND`-env dev logic; **root workspace `package.json`** so `npm run dev` / `npm run build` run from the example root.
- **Motoko:** moc 1.11.0 + `--default-persistent-actors` (dropped `persistent`); committed `backend/backend.did`.
- **Rust:** unchanged; relocated; the `.did` renamed to `backend/backend.did` and the Makefile updated to match.
- README split per language (linked prerequisites); `test.sh` added (query method called with `--query`); `vetkeys-password-manager.yml` split into `motoko` + `rust` jobs.

## CODEOWNERS

## Verification (local)
- **Motoko:** `vite build` + deploy, `test.sh` PASS, and an inserted value **survives an upgrade**.

> Part of the vetkeys split series (#1443, #1444, #1445). Last remaining: `password_manager_with_metadata`.


marc0olo:
Thanks @Copilot. I checked each item against the original shared frontend. Every flagged file is byte-identical to the pre-split version — this PR only splits the example into self-contained Motoko/Rust copies (import paths, the `PUBLIC_CANISTER_ID:backend` env key, vite/bindgen wiring). None of these were introduced here, so to keep the split faithful and avoid re-testing behavior that already works, I'm leaving them untouched. They're better addressed upstream:

Pre-existing frontend items:
- **`store/vaults.ts` — uninitialized `vaultPollerHandle`**: harmless; `clearInterval(undefined)` is a no-op.
- **`store/draft.ts` — draft restore checks `content` && `tags`**: pre-existing; draft restore is effectively disabled.
- **`components/NewPassword.svelte` / `SharingEditor.svelte` — `Principal.fromText` can throw outside the try/catch**: pre-existing input-validation gap.
- **`components/EditPassword.svelte` (move error handling)**: pre-existing; sets `deleting` instead of clearing `updating`.
- **`components/SharingEditor.svelte` filter with `.compareTo()`**: this is a real pre-existing bug — `.compareTo()` returns a string, which is always truthy, so the filter keeps all users. Worth fixing upstream.
- **`components/EditPassword.svelte` class ternary**: pre-existing and malformed, but a minor correction — it does not fail to compile. Svelte parses `{'Read' in accessRights}` as a complete interpolation and treats the trailing `? 'hidden' : ''}` as literal text, so the class renders incorrectly rather than breaking the build.

Other:
- **`frontend/README.md` (dfx / `encrypted_maps_example` / old env var)**: pre-existing and now stale relative to the icp-based setup; carried over unchanged.
- **CI has no `test.sh`**: intentional. None of these vetKeys examples shipped a `test.sh` on `master`, and CI's default identity is anonymous (`2vxsx-fae`), which several of these backends reject — a CLI `test.sh` would trap in CI even though it passes locally. The workflow is deploy-only, which still verifies build + install.

marc0olo:
Follow-up: I removed `frontend/README.md` from both the Motoko and Rust copies (697d5971). It was carried over from the original shared frontend and was the only `frontend/README.md` in the repo — no other example ships one — and it referenced dfx, an `encrypted_maps_example` canister, and the old `CANISTER_ID_IC_VETKEYS_ENCRYPTED_MAPS_CANISTER` env var, all contradicting the icp-based top-level README. Rather than maintain duplicate, misleading instructions, it's dropped.

On the CI point: the workflow already deploys both backends from their own working directories in separate jobs (`motoko/vetkeys/password_manager` and `rust/vetkeys/password_manager`), and both jobs pass. It stays deploy-only (build + install + asset sync) by design, consistent with the other vetKeys examples.

raymondk (approved):

pierugo-dfinity (approved):

PR #1447, opened 2026-07-16
chore(vetkeys): split password_manager_with_metadata into self-contained motoko + rust examples

marc0olo:
## Summary

Splits the vetkeys **`password_manager_with_metadata`** example into two self-contained examples (one per language) — same restructure as #1443–#1446, and the **last** of the vetkeys split series. Previously it lived only under `rust/vetkeys/password_manager_with_metadata/` with a shared **Svelte** frontend symlinked into `motoko/` and `rust/` subdirs.

```
motoko/vetkeys/password_manager_with_metadata/   backend/app.mo + own frontend/ + icp.yaml + mops.toml
rust/vetkeys/password_manager_with_metadata/      backend/ (flattened) + own frontend/ + icp.yaml + Cargo.toml
```

## Why this one is a hybrid (bindgen, but no camelCase)

This example has a **custom backend** (metadata methods) that the frontend calls via **generated bindings** (`password_manager.ts` builds an actor from `idlFactory`), *plus* the **`@icp-sdk/vetkeys` Encrypted Maps client** for the crypto ops. Because the SDK client calls the Encrypted Maps interface **snake_case**, the backend interface is kept snake_case (**no camelCase**), while the custom actor still needs **icpBindgen**.

## Changes
- **Frontend (Svelte):** per-language copy (identical); bindings via the **`@icp-sdk/bindgen` Vite plugin** from the committed `backend/backend.did` (replaces `scripts/gen_bindings.sh`); declaration imports repointed to `../bindings/…`; canister renamed to **`backend`** (env key `PUBLIC_CANISTER_ID:backend`); dropped eslint/prettier and a stray `save` dependency; kept the `@dfinity/vetkeys` → `@icp-sdk/vetkeys` vite alias; **root workspace `package.json`** so `npm run dev` runs from the example root.
- **Motoko:** moc 1.11.0 + `--default-persistent-actors` (dropped `persistent`); committed `backend/backend.did`.
- **Rust:** unchanged; relocated (the `.did` was already `backend/backend.did`).
- README split per language (linked prerequisites); `test.sh` added (query method called with `--query`); `vetkeys-password-manager-with-metadata.yml` split into `motoko` + `rust` jobs.

## CODEOWNERS

## Verification (local)
- **Motoko:** `vite build` (icpBindgen) + deploy, `test.sh` PASS, and a value **survives an upgrade**.

> Completes the vetkeys split series: #1443 (basic_ibe), #1444 (basic_bls_signing), #1445 (encrypted_notes_app_vetkd), #1446 (password_manager), and this one.


marc0olo:
Thanks @Copilot. I checked all 11 items against the original shared frontend: every flagged file is byte-identical to the pre-split version. This PR only splits the example into self-contained Motoko/Rust copies (import paths, the `PUBLIC_CANISTER_ID:backend` env key, vite/bindgen wiring), so none of these were introduced here. To keep the split faithful and avoid re-testing behavior that already works, I'm leaving them untouched — they're best addressed upstream.

Genuine pre-existing bugs worth an upstream fix:
- **`SharingEditor.svelte` filter with `.compareTo()`**: returns a string (always truthy), so the target user is never removed.
- **`SharingEditor.svelte` vault lookup (`===` + `vaultOwnewr` typo)**: compares Principals by reference; can fail for equal values.
- **`Password.svelte` default `metadata: undefined`** vs required field: typing inconsistency.
- **`EditVault.svelte` `users.find(...)` + `in`**: can throw when the current user isn't in `vault.users`.
- **`Passwords.svelte` copy** ("Filter notes…" / empty-state) references notes instead of passwords.
- **`Vault.svelte` class string** with stray characters: Tailwind classes may not apply.

Minor corrections:
- **`vaults.ts` uninitialized `vaultPollerHandle`**: harmless; `clearInterval(undefined)` is a no-op.
- **`draft.ts` restore checks `tags`**: pre-existing; draft restore is effectively disabled.
- **`EditPassword.svelte` class ternary**: pre-existing and malformed, but it does **not** fail compilation — Svelte parses `{'Read' in accessRights}` as a complete interpolation and treats the trailing `? 'hidden' : ''}` as literal text. The class renders incorrectly rather than breaking the build (this PR's `motoko` and `rust` CI jobs both build the frontend and pass).

CI:
- **No `test.sh`**: intentional. None of these vetKeys examples shipped a `test.sh` on `master`, and CI's default identity is anonymous (`2vxsx-fae`). The workflow deploys **both** backends from their own working directories (`motoko/...` and `rust/...`) in separate jobs — both pass — which verifies build + install + asset sync.

pierugo-dfinity (approved):

raymondk (approved):

PR #1448, opened 2026-07-22
ci: centralize dev-env image version + bump to v1.2.0

marc0olo:
## What

Centralizes the `icp-dev-env` CI image version into a single source, bumps it to **`v1.2.0`**, and fixes a few image-variant mismatches. Previously the version was hardcoded across ~60 workflow files plus the devcontainer, which had already drifted (devcontainer was on `0.3.1` while CI was on `1.0.1`).

## Changes (net, vs `master`)

- **New `.github/workflows/_run-example.yml`** — a reusable (`workflow_call`) workflow that runs an example's CI steps inside the dev-env container. It is the single source of truth for the CI image version. Callers pass `language` (image variant), `working-directory`, and `run` (commands). Examples needing a PocketIC server use the optional `install-pocketic` / `pocketic-version` inputs.
- **All example workflows converted to thin callers** of `_run-example.yml` — no workflow hardcodes the image version anymore. Extra setup steps (apt installs, `rustup target add`, model download) are folded into `run`; PocketIC integration tests use `install-pocketic`.
- **New `.github/workflows/dev-env-version-check.yml`** — fails the PR if the devcontainer and the reusable workflow drift to different versions.
- **`.devcontainer/devcontainer.json`** bumped `icp-dev-env-all:0.3.1` → `v1.2.0`.
- **Image-variant fixes**: frontend-only `hosting/react` and `hosting/oisy-signer-demo` now use the `all` image (were incorrectly on `motoko`), consistent with the other `hosting/*` examples.
- **`.github/workflow-template.yml`** updated to the caller pattern.

### Tag convention (`v`-prefix)

The dev-env release CI tags each release with a `v` prefix and publishes **both** forms to the registry — `type=ref,event=tag` yields `v1.2.0`, `type=semver,pattern={{version}}` yields `1.2.0`. We pin the **`v`-prefixed** form (`v1.2.0`) to match the git release tag. `dev-env-version-check.yml` extracts the numeric `X.Y.Z` from both files, so the prefix does not affect the drift check.

Not touched: `basic_bitcoin` (bespoke workflow — needs a bitcoind regtest node and Docker bind-mounts, so it can't use the shared container-based workflow) and `ninja_pr_checks` (uses the unrelated `-slim` image on a Node-version tag scheme).

## Result

Bumping the dev-env version is now a one-line change in `_run-example.yml` + `devcontainer.json` (kept equal by `dev-env-version-check`). Because each caller watches `_run-example.yml` in its `paths:`, a bump automatically re-runs every affected example — which is what this PR itself does: all example suites run against `v1.2.0`.

## Validation

The `v1.2.0` images are published (via dfinity/icp-dev-env#43), and **all 63 checks pass** on this PR: every example suite (Motoko + Rust, including the PocketIC and ML examples — `image-classification`, `face-recognition`), plus the `dev-env-version-check` drift guard with both places on `v1.2.0`.


marc0olo:
Thanks @Copilot — both points were fair and are addressed in 315dbbb9:

- **`workflow-template.yml` paths**: the template now shows the two-language (Motoko + Rust) form by default, with one `pull_request` path entry per language directory, and documents how to adapt it for single-language, frontend-only (`language: all` + `hosting/<name>`), and PocketIC cases. This removes the earlier mismatch where a copied two-language workflow would only trigger on one language directory.
- **`dev-env-version-check.yml` grep**: the version is now extracted from the actual `container: ghcr.io/dfinity/icp-dev-env` line (first match, anchored to the key), so a comment or example line can't be matched by mistake. Verified it still resolves `1.1.0` on both sides.

lwshang (approved):

PR #1454, opened 2026-07-28
refactor(motoko/icrc2-swap): import the ICRC .did directly (moc 1.13.0 `idl:`)

marc0olo:
## What

Types `icrc2-swap`'s token-ledger calls against the ICRC-1/ICRC-2 **Candid interface imported directly** via Motoko's `idl:` import (moc 1.13.0+), replacing the hand-written `backend/ICRC.mo` type module. No bindings are generated or committed, and no extra tooling (`didc`) is needed.

```motoko
import ICRC "idl:../candid/icrc.did";
```

> **Update:** this PR originally committed a `didc`-generated `bindings/ICRC.mo`. moc **1.13.0** ([release](https://github.com/caffeinelabs/motoko/releases/tag/1.13.0), [docs](https://docs.internetcomputer.org/languages/motoko/fundamentals/modules-imports/#importing-a-local-did-file)) now imports a `.did` directly, so the generated bindings and the `didc` step are gone — only the committed `candid/icrc.did` remains.

## Why this pattern for ICRC

`icrc2-swap` holds two ledgers at once (`token_a`, `token_b`) and picks the target **per call** (`actor(args.token.toText()) : ICRC.Self`). That's the "one interface, many ledgers, runtime-chosen principal" case, which `--actor-env-alias`/`--actor-id-alias` (which bind one import to one principal) can't express. The `idl:` import gives **types only** (`ICRC.Self` + named types like `ICRC.Account`, `ICRC.TransferArgs`); the backend supplies its own `actor(principal)` per token. This generalizes to any DeFi-shaped backend that talks to many ICRC ledgers (ICP, ckBTC/ckETH, SNS tokens, …).

This completes the inter-canister taxonomy across the Motoko examples:

| Situation | Mechanism | Example |
|---|---|---|
| id fixed & universal | `--actor-id-alias` | icp_transfer (#1455) |
| id varies per environment, injected | `--actor-env-alias` | parallel_calls, pub-sub, ic-pos, evm_block_explorer, llm_chatbot |
| id chosen at runtime / many targets | **`idl:` import + `actor(id)`** | **icrc2-swap** (this PR) |

## Changes

- **`candid/icrc.did`** — the ICRC-1/ICRC-2 interface (focused plain service; types from the `ic-icrc1-ledger` reference wasm). Committed source of truth, imported directly.
- **`backend/app.mo`** — `import ICRC "idl:../candid/icrc.did"`; types are `ICRC.Self`/`ICRC.Account`/`ICRC.TransferArgs`/…; hand-written `ICRC.mo` removed.
- **`mops.toml`** — `moc` 1.9.0 → **1.13.0** (required for the `idl:` import); `mo:core` 2.5.0 → 2.6.0.
- Unrelated tidy-up in the same example: Map `swap` (return always discarded) → `add`.
- README updated: the interface is imported directly; nothing to regenerate, no `didc`.

## On tooling (cc @Kamirus)

Earlier revisions of this PR leaned on `didc bind` + a committed `bindings/ICRC.mo`, and flagged [dfinity/candid#729](https://github.com/dfinity/candid/pull/729) (PascalCase Motoko output) plus the idea of `mops` generating bindings as a pre-build step. moc 1.13.0's native `.did` import largely supersedes that for this use case: the compiler reads the `.did` directly, applies the PascalCase convention itself, and there's nothing to generate, commit, or install. Only the committed `.did` remains — which is exactly the "commit the interface, not the generated code" end state we were aiming for.

## Testing

`bash deploy.sh && bash test.sh` — **all 7 tests pass** (deposit via `icrc2_transfer_from`, atomic swap, withdraw via `icrc1_transfer` with correct fee-adjusted balance deltas), confirming the `idl:`-imported types are wire-compatible with the real ledger.


Kamirus:
Yes, I think once https://github.com/dfinity/candid/pull/729 lands it would make sense for mops to adopt this tool.
Mops would handle fetching it via `mops toolchain` and we could wrap it with `mops generate ...` command

raymondk (approved):

PR #1469, opened 2026-08-06
chore: switch ICP skills from pinned to autosync

marc0olo:
Switches the repo's ICP-skills setup from **pinned** (`skills-lock.json` + `npx skills update`) to **autosync** (a SessionStart hook that mirrors the latest skills each session). Uses the mechanism from [dfinity/icskills#251](https://github.com/dfinity/icskills/pull/251), hardened by [#257](https://github.com/dfinity/icskills/pull/257).

## Why

The goal is to **never commit anything when a skill changes**. Pinned mode fails that — every refresh rewrites `skills-lock.json`, which has to be committed. Autosync commits the hook + script **once**; skills then refresh with zero commits, always current.

- **Robust mechanism**: the sync is a differential, hash-keyed mirror — it re-downloads only changed skills, prunes removed ones, swaps each changed skill's dir atomically (so intra-skill file renames don't orphan), and degrades gracefully if offline or `jq` is missing.
- **Matches repo philosophy**: everything here aligns examples with the fast-moving platform; always-latest skills fit better than a manually-refreshed pin.
- **No reproducibility loss that matters**: skills guide agents authoring examples; they are not build inputs (CI runs in containers and never reads them).

## What changed

- **Add** `.claude/sync-ic-skills.sh` (fetched verbatim from the published autosync skill) + `.claude/settings.json` with a `SessionStart` hook — both committed so the whole team gets auto-updating skills.
- **`.gitignore`**: ignore `.claude/skills/` (the managed cache) and `.claude/settings.local.json` (personal); drop the old `.agents/skills` line.
- **Remove** `skills-lock.json`.
- **AGENTS.md**: managed block → `autosync` state; added an **on-demand fallback** note so non–Claude Code agents (Cursor, Copilot, Codex) still know how to fetch skills, since the hook is Claude Code-only.

## Tradeoffs (called out for review)

- **Claude Code-only hook** — mitigated by the documented on-demand fallback for other agents.
- **A committed `SessionStart` hook runs a network-fetching script.** Anyone opening the repo root in Claude Code gets a one-time trust prompt for `.claude/sync-ic-skills.sh`. It's from the official `skills.internetcomputer.org` and only writes to `.claude/skills/`. Acceptable given the audience is the DX team (repo isn't taking external contributions), but flagging it explicitly.

Verified locally: hook runs, 26 skills synced into `.claude/skills/`, re-run is a silent no-op, no leftover staging dirs.


lwshang (approved):

marc0olo:
@raymondk I will merge this now. once the committed autosync script starts to become an issue when drifting away I will think about this topic again.

PR #1472, opened 2026-08-06
Stop committing lock files; add weekly CI safety net

### LINKED ISSUE #1457: Inconsistent Cargo.lock handling across Rust examples — pick one policy
## Problem

The repo currently has a three-way inconsistency around `Cargo.lock`:

- The root `.gitignore` ignores `Cargo.lock`.
- **30 examples commit it anyway** (tracked files override the ignore): `rust/backend_wasm64`, `rust/basic_bitcoin`, `rust/basic_ethereum`, `rust/candid_type_generation`, `rust/canister-info`, `rust/canister-snapshot-download`, `rust/canister-snapshots`, `rust/canister_logs`, `rust/exchange-rates`, `rust/face-recognition`, `rust/guards`, `rust/icp_transfer`, `rust/image-classification`, `rust/inter-canister-calls`, `rust/low_wasm_memory`, `rust/parallel_calls`, `rust/performance_counters`, `rust/periodic_tasks`, `rust/photo_gallery`, `rust/qrcode`, `rust/query_stats`, `rust/receiving-icp`, `rust/send_http_get`, `rust/send_http_post`, `rust/simd`, `rust/threshold-schnorr`, `rust/unit_testable_rust_canister`, `rust/who_am_i`, `rust/x509`, and `motoko/parallel_calls` (Rust load-test tool).
- The remaining Rust examples (`rust/hello_world`, `rust/vetkeys/*`, `rust/evm_block_explorer`, `rust/llm_chatbot`, `rust/threshold-ecdsa`, ...) do not commit it.

## Current impact: none of the committed locks are used by builds

- The `@dfinity/rust` recipe ([icp-cli-recipes](https://github.com/dfinity/icp-cli-recipes/tree/main/recipes/rust)) has a `locked: boolean` option that appends `--locked` to `cargo build` — but it **defaults to false and no example sets it**.
- Three examples use custom `cargo build` steps in `icp.yaml` (`image-classification`, `face-recognition`, `composite_query`); none pass `--locked`/`--frozen`.
- The only `--locked` flags in the repo are `cargo install --locked` for CI tools (`candid-extractor`, `wasi2ic`), which use the tool's own bundled lock file, not the example's.

So today the committed locks only pin what developers get locally; CI always resolves fresh dependencies.

## Options

**(a) Commit `Cargo.lock` everywhere and make it load-bearing**
- Remove `Cargo.lock` from the root `.gitignore`, commit the missing locks, set `locked: true` in every rust recipe config (and add `--locked` to the three custom build steps).
- Pro: reproducible builds; local dev matches CI.
- Con: locks go stale unless updates are automated (Renovate/Dependabot); more churn in PRs.

**(b) Untrack all locks and stay lock-free**
- `git rm --cached` the 30 tracked locks; keep the root gitignore entry.
- Pro: zero maintenance; CI always builds against the latest compatible dependencies, catching upstream breakage early — arguably the right behavior for an examples repo that wants to stay current.
- Con: builds are not reproducible; a bad upstream release can break CI for unrelated PRs.

Either option is better than the current mixed state.

marc0olo:
Closes #1457. Resolves the inconsistent lock-file handling and makes examples resolve current dependencies on clone.

## The problem (verified against `master`)
- **30 `Cargo.lock` + 1 `mops.lock`** are committed, yet **both are gitignored** — the tracked ones are overrides, so the policy is inconsistent.
- **No build depends on them**: nothing sets `locked: true` in any `icp.yaml`, and there's no `--locked`/`--frozen` anywhere except the CI tool bootstrap (`cargo install --locked` for `candid-extractor`/`wasi2ic`, which use the tool's own lock).

## Decision: don't commit lock files
A developer cloning an example should get the **latest semver-compatible** dependencies (per each `Cargo.toml`/`mops.toml` range), not a stale pin — matching the repo's stay-current philosophy. Since the locks aren't load-bearing, removing them changes nothing about how builds run today; it only lets resolution pick current versions.

- Untracked all 30 `Cargo.lock` and the 1 `mops.lock` (`git rm --cached`; both already in `.gitignore`, so they stay ignored). `package-lock.json` was already gitignored.

## Safety net: weekly CI
Without locks, a bad upstream patch could break an example with no repo change — and only surface when someone touches it. So every example workflow (and `workflow-template.yml`) gains a weekly `schedule` trigger (`cron: "30 6 * * 1"`), building/deploying/testing all examples every Monday.

- Chose per-workflow schedule (not a new aggregate workflow) so each run reuses its example's **exact** config — pocketic, `deploy.sh`, cycles, language — with no duplication or drift, and new examples inherit it from the template.

AGENTS.md documents both the no-lock-files policy and the weekly run.

## Notes
- All 63 modified workflows validate as YAML.
- Removing the 30 locks means those examples' next CI run resolves fresh deps for the first time — if anything was silently pinned to an old transitive version, the weekly run (or this PR's CI) will surface it. That's the intended signal.


mbjorkqvist (commented):
Thanks for looking into this, @marc0olo! Looking at [the referenced issue #1457](https://github.com/dfinity/examples/issues/1457), I'm curious why you didn't go for option a, with e.g., a weekly dependabot run? What happens if the weekly CI check now fails, would that require manual intervention to e.g., fix an example to work with the latest version of some dependency? Maybe I'm missing something, e.g., how things may be significantly different in this repo with lots of examples, a mix of Rust and Motoko, and `icp-cli`?

marc0olo:
@mbjorkqvist my personal take on this is mainly that we would have a lot of noise on the repo and generally we want everybody who clones and runs the examples to automatically resolve the latest patch versions which isn't the case if we define strict versions in the lock files.

either way, our goal should be to keep the dependencies on this repo up to date. we need to figure out the best way to handle that efficiently.

> would that require manual intervention to e.g., fix an example to work with the latest version of some dependency?

we will see, but ideally the CI never fails for the examples (although I recognized some flaky tests already which I still need to look at, they usually pass on a 2nd run). for the UI we don't have integration tests. but at least compilation + deployment and in many cases also backend functionality is tested via CI.

of course we should aim to avoid manual intervention. (e.g. dependabot PRs would also have to be handled by "someone", whether it is an agent or a human)

wdyt?

mbjorkqvist (approved):

lwshang (approved):

pierugo-dfinity (approved):

yhabib (approved):

PR #1474, opened 2026-08-12
feat(vetkeys): adopt custom_value_endpoints in password_manager_with_metadata (Rust)

marc0olo:
Adopts the `custom_value_endpoints` form of `export_encrypted_maps_canister!`, shipped in [`ic-vetkeys` 0.9.0](https://crates.io/crates/ic-vetkeys/0.9.0) and closing [dfinity/vetkeys#423](https://github.com/dfinity/vetkeys/issues/423) (PRs dfinity/vetkeys#424, #425).

## What the new macro form gives us

It generates the stable state, the `#[init]`/`#[post_upgrade]` hooks, the 8 control-plane endpoints (vetKD keys, access control, map-name enumeration) and the `with_encrypted_maps`/`with_encrypted_maps_mut` accessors — but **none** of the 7 endpoints that read or write encrypted values.

That is precisely what `password_manager_with_metadata` needs. It keeps a backend-authoritative metadata row (creation date, modification counter, tags, url) per encrypted value, so the raw `insert_encrypted_value` / `remove_encrypted_value` mutators must not be exposed: they would write a value with no metadata row and desync the two stores. Until 0.9.0 the generator was all-or-nothing, so the example hand-wrote ~11 endpoints — 8 of them character-for-character what the macro emits.

## Changes

**`password_manager_with_metadata` backend: 343 → 193 lines.** Deleted the `ENCRYPTED_MAPS` thread-local, `init`/`post_upgrade`/`init_encrypted_maps`, the `KEY_NAME` `StableCell` at `MemoryId(5)`, and the 8 passthrough endpoints. What remains is the genuinely app-specific part: `PasswordMetadata`, the metadata map at `MemoryId(4)`, and the three `*_with_metadata` endpoints written against the accessors.

The key-name cell was redundant: `KeyManager::init` persists the domain separator and vetKD key id in its config cell, and `StableCell::init` ignores the passed defaults once the cell holds a value — so the macro's `#[post_upgrade]` re-attaches from the persisted config. Verified on a live upgrade: `get_vetkey_verification_key` returns byte-identical output before and after.

**Hardened the metadata read path.** `get_encrypted_values_for_map_with_metadata` paired values with metadata by zipping two ordered iterators, guarded only by a `debug_assert_eq!` that compiles out in release — a divergence would have silently returned a password with someone else's metadata. It now looks metadata up per key and reports a missing row as an error instead.

**Domain separators renamed to match their examples.** The metadata example used `password_manager_app` (reading as if it belonged to the plain example) and the plain example used the library-flavoured `encrypted_maps_app`; they are now `password_manager_with_metadata_app` and `password_manager_app`. Note the separator is persisted in the config cell, so an upgraded canister keeps its old value — only fresh installs pick up the new names. Harmless either way: the separator is a vetKD derivation context combined with the calling canister's ID, so it was never a correctness issue, only a naming one.

**`ic-vetkeys` 0.8.1 → 0.9.0 in all five Rust examples that pin it.** The release is additive and the full macro form is unchanged, so `basic_ibe`, `basic_bls_signing` and `basic_timelock_ibe` are pin-only changes.

**README.** The backend note claimed the macro "offers no way to override or omit" the mutators, which 0.9.0 fixed; rewritten to explain when to reach for each form.

## Verification

- All five examples build for `wasm32-unknown-unknown`; `cargo clippy -- -Dwarnings` clean.
- **`backend.did` byte-identical** for every example with committed candid (`password_manager_with_metadata`, `password_manager`, `basic_ibe`, `basic_bls_signing`), so no frontend and no `icp.yaml` change — the generated `#[init]` still takes the vetKD key name.
- Local `icp deploy` end-to-end on both password managers: create/read/update/remove with metadata, vault sharing to a second Internet Identity, decryption of existing passwords after an upgrade (including the persisted `IndexedDbDerivedKeyMaterialCache`), and `insert_encrypted_value` correctly absent from the metadata canister (`IC0536: Canister has no update method`).

## Follow-up

Motoko is untouched here. `ic-vetkeys` 0.6.0 for mops ships the mirrored `EncryptedMapsControlPlaneCanister` mixin, but adopting it also requires `moc` 1.13.0 + `mo:core` 2.6.1 and moves key configuration to the `VETKD_KEY_NAME` canister environment variable — a larger change that deserves its own PR. Both Motoko password manager examples also currently share the same `password_manager_example_app` separator, to be fixed in that pass.


rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs
```diff
@@ -1,12 +1,22 @@
...
+// canister — not the client — maintains.
+//
+// The `custom_value_endpoints` form of the library macro generates the state,
+// the `#[init]`/`#[post_upgrade]`, the control-plane endpoints (vetKD keys,
+// access control, map-name enumeration) and the `with_encrypted_maps`/`_mut`
+// accessors, but *no* endpoints that read or write encrypted values. That is
+// what this canister needs: the raw `insert_encrypted_value` /
+// `remove_encrypted_value` mutators are never exposed, so nothing can write a
+// value without its metadata row, and the `*_with_metadata` endpoints below
+// update both stores in a single call.
```
alin-at-dfinity: Nit: Maybe make this a module doc comment?


```suggestion
//! This canister layers a second, backend-authoritative store on top of the
//! `ic-vetkeys` EncryptedMaps library: every encrypted password has a matching
//! metadata row (creation date, modification counter, tags, url) that the
//! canister — not the client — maintains.
//!
//! The `custom_value_endpoints` form of the library macro generates the state,
//! the `#[init]`/`#[post_upgrade]`, the control-plane endpoints (vetKD keys,
//! access control, map-name enumeration) and the `with_encrypted_maps`/`_mut`
//! accessors, but *no* endpoints that read or write encrypted values. That is
//! what this canister needs: the raw `insert_encrypted_value` /
//! `remove_encrypted_value` mutators are never exposed, so nothing can write a
//! value without its metadata row, and the `*_with_metadata` endpoints below
//! update both stores in a single call.
```

rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs
```diff
@@ -74,133 +84,64 @@ type StableMetadataMap = StableBTreeMap<(MapOwner, MapName, MapKey), PasswordMet
...
-}
+// The first argument is the domain-separator string that isolates this
+// application's derived keys; it must stay stable for the life of the canister.
+// The four Memory instances back EncryptedMaps' own stable state, in the order
+// it expects: config (which persists the domain separator and vetKD key id),
+// access control, shared keys, and the encrypted values. This canister's own
+// metadata map lives in the same MemoryManager under id 4.
+//
+// The generated `#[init]` takes the vetKD key name (see `init_args` in
+// `icp.yaml`); `#[post_upgrade]` re-reads it from the persisted config.
```
alin-at-dfinity, line 105: Similarly, this may (or may not, I'm not entirely sure of the context) work as a doc comment:

```suggestion
/// The first argument is the domain-separator string that isolates this
/// application's derived keys; it must stay stable for the life of the canister.
/// The four Memory instances back EncryptedMaps' own stable state, in the order
/// it expects: config (which persists the domain separator and vetKD key id),
/// access control, shared keys, and the encrypted values. This canister's own
/// metadata map lives in the same MemoryManager under id 4.
///
/// The generated `#[init]` takes the vetKD key name (see `init_args` in
/// `icp.yaml`); `#[post_upgrade]` re-reads it from the persisted config.
```

alin-at-dfinity (approved):
A couple of suggestions, otherwise LGTM.

rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs
marc0olo: Done in f3e9fdb1 — applied to both password managers so the pair stays consistent.

rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs
marc0olo: Good instinct to hedge — this one doesn't work. A `///` here triggers rustc's `unused_doc_comment`:

```
warning: unused doc comment
  --> backend/src/lib.rs:97:1
   |
97 | / /// The first argument is the domain-separator string that isolates this
   | |_...
   = note: rustdoc does not generate documentation for macro invocations
```

The macro expands to the endpoints themselves, so there's no item for the doc to attach to. Left as a plain `//` comment.

Issue #1459, opened 2026-07-30
hosting/photo-storage broken on static-site recipe — revert to asset-canister interim, decide long-term fate


marc0olo:
Part of #1458.

`hosting/photo-storage` demonstrates programmatic uploads via `AssetManager` from `@icp-sdk/canisters/assets`. Since #1449 switched it to `@dfinity/static-site`, the example is broken at runtime: the certified-assets canister exposes none of the legacy API (`list`, `store`, `create_batch`, `commit_batch` — verified against [certified-assets.did](https://github.com/dfinity/certified-assets/blob/main/certified-assets.did)). `App.jsx` calls `assetManager.list()` on page load, so the app fails immediately. Only `authorize` coincidentally still resolves (certified-assets has a flat authorize model), masking the breakage.

Rewriting against the new interface is not viable: certified-assets mutations go through a single-session `start_sync`/`upload_chunks`/`execute_operations` protocol that is version-locked to the recipe's sync plugin and designed for deploys, not concurrent end-user uploads.

**Interim fix (unblocks the example now):** revert the recipe to `@dfinity/asset-canister@v2.2.1` — the only canister supporting the AssetManager flow — with a comment explaining the deliberate exception, and verify the app works end-to-end again.

**Long-term decision:**
- (a) keep it permanently on the legacy recipe as *the* programmatic-uploads example, clearly labeled;
- (b) repurpose it to store photos in a purpose-built backend canister (overlaps with `rust/photo_gallery`);
- (c) retire it.

Depends on whether the legacy asset canister remains supported ([icskills#256](https://github.com/dfinity/icskills/pull/256) keeps it as documented legacy) and whether a runtime-upload story exists/is planned for certified-assets.

marc0olo:
Status: the interim revert is applied locally and will land with the next batch of commits:

- `icp.yaml` recipe reverted to `@dfinity/asset-canister@v2.2.1` with a comment explaining the deliberate exception (so the next repo-wide recipe sweep doesn't silently re-migrate it).
- Bonus fix found on the way: the README's authorize command targeted a canister named `photo-storage`, but the canister is named `frontend` in `icp.yaml` — the documented command could never have worked. Corrected to `icp canister call frontend authorize ...`.

Remaining in this issue:
- [ ] Manual e2e verification of the reverted example (deploy, authorize the hardcoded identity, upload a photo, reload → photo persists)
[...]

Issue #1458, opened 2026-07-30
Incomplete static-site (certified-assets) migration: broken AssetManager example, dead .ic-assets.json5 configs, lost security headers


marc0olo:
All frontends in this repo were switched to the `@dfinity/static-site` recipe (the **certified-assets** canister). That recipe swap was only step 1 of the migration: `@dfinity/asset-canister` and `@dfinity/static-site` are **different canisters with different APIs and config mechanisms**, not a rename. Audited against the migration requirements in [dfinity/icskills#256](https://github.com/dfinity/icskills/pull/256) (static-site skill + `references/migrating-from-asset-canister.md`), the following gaps remain.

## 1. `hosting/photo-storage` is functionally broken

The example's core purpose — programmatic uploads via `AssetManager` from `@icp-sdk/canisters/assets` — cannot work against certified-assets, which does not expose the legacy `store`/`create_batch`/`commit_batch`/`list` API (confirmed by icskills#256, eval 3: "AssetManager doesn't work on static-site").

- `src/App.jsx` calls `assetManager.list()` (L56) and `batch.store(...)` (L75) → method-not-found at runtime.
- The README's `authorize` instruction coincidentally still resolves (certified-assets has a flat `authorize(principal)` model), which masks the breakage until an upload is attempted.

**Decision needed:** either (a) rework the example so photos are stored via a purpose-built backend canister (like `rust/photo_gallery`), (b) keep it deliberately on the legacy `@dfinity/asset-canister` recipe as a "programmatic uploads" example with a clear note, or (c) retire it. Uploads to certified-assets go exclusively through the recipe's sync plugin on `icp deploy`.

## 2. 18 dead `.ic-assets.json5` files — including silently lost security headers

These files are ignored by certified-assets, yet still sit in `public/` (or `assets/`), so they are **uploaded and publicly served as regular assets**:

`hosting/oisy-signer-demo`, `motoko/ic-pos`, `motoko/vetkeys/{basic_bls_signing, basic_ibe, basic_vetkd, encrypted_notes_app_vetkd, password_manager, password_manager_with_metadata}`, `rust/image-classification`, `rust/photo_gallery`, `rust/qrcode`, `rust/vetkeys/{basic_bls_signing, basic_ibe, basic_timelock_ibe, basic_vetkd, encrypted_notes_app_vetkd, password_manager, password_manager_with_metadata}`

The real damage is what these configs *used to provide* and certified-assets does **not** provide by default (it ships no default security headers):

[...]

PR #1478, opened 2026-08-27
chore(hosting): drop the photo-storage example

marc0olo:
Retires `hosting/photo-storage`.

The example demonstrated programmatic, end-user-facing uploads via `AssetManager` from `@icp-sdk/canisters/assets` — an API only the legacy asset canister provides. We no longer want to advertise the asset canister as programmatic storage, so the example is **retired** rather than reworked or kept on the legacy recipe.

This settles the long-term decision left open in #1459 (option **(c) retire**) and resolves item 1 of the umbrella #1458.

### Why not the other options

- **Rework onto certified-assets** — not viable: it exposes no runtime-upload API. Mutations go through a single-session `start_sync`/`upload_chunks`/`execute_operations` protocol that is version-locked to the recipe's sync plugin and designed for deploys, not concurrent end-user uploads.
- **Repurpose onto a purpose-built backend canister** — duplicates `rust/photo_gallery`, which already does exactly that.
- **Keep it on the legacy recipe** — it was the only example still pinned to `@dfinity/asset-canister`, blocking a clean "all frontends are certified-assets" story.

### What's removed

- `hosting/photo-storage/` (22 files)
- `.github/workflows/hosting-photo-storage-example.yml`
- its `CODEOWNERS` entry
- both `hosting/README.md` entries, including the "one deliberate exception" note about the legacy recipe
- the `photo-storage` reference in the AGENTS.md ICP Ninja eligibility rules — genericized to "an `authorize` call that only a controller can make"; the controller-only-setup rule itself still stands

### Result

Every frontend in the repo now runs on `@dfinity/static-site` (certified-assets). `grep` for `photo-storage` comes back empty (`rust/photo_gallery` is a different example and is untouched).

Docs/README fixes surfaced while auditing this are split out into a separate PR — no file overlap, the two can merge in either order.


marc0olo:
Added a second commit (4d10904) after auditing how this repo refers to the retired asset canister.

**Where its source actually is** — `dfinity/sdk/src/canisters/frontend/`:

| Crate | What it is |
|---|---|
| `ic-frontend-canister/` | the wasm — README: *"This is the implementation of the DFX assets canister in Rust with support for asset certification"* |
| `ic-certified-assets/` (v0.3.0) | the library it wraps |
| `ic-asset/` | upload/sync library; holds `security_policy.rs`, the authoritative legacy header policy |
| `icx-asset/` | standalone CLI |

`dfinity/sdk` is not archived, so nothing on the repo signals that this is retired.

**The trap:** the legacy crate is named `ic-certified-assets`, while the canister we actually use is the separate `dfinity/certified-assets` repo. An agent told "use certified-assets, not the legacy asset canister" can search GitHub, land in the sdk crate, and read the wrong source with full confidence. AGENTS.md now names both locations at the point where `.ic-assets.json5` is banned — that felt like the one fact worth spending a sentence on.

**Dropped the docs link in `hosting/README.md`.** It pointed "asset canister" at `docs.internetcomputer.org/guides/frontends/asset-canister`. That page is still live (200) but wholly legacy — it teaches `.ic-assets.json5` SPA routing, AssetManager programmatic uploads, and `http://<canister-id>.localhost:8000` URLs, i.e. exactly the three things these examples no longer do. The orienting clause stays ("the successor to the dfx-era asset canister"), just without a link sending readers to a page that contradicts every example it introduces.

Follow-up for a different repo, not this PR: `developer-docs/docs/guides/frontends/asset-canister.md` (329 lines, `sidebar order: 1`) wants a rewrite around certified-assets + the static-site recipe. The `static-site` skill there already has `references/legacy-asset-canister.md` and `migrating-from-asset-canister.md`, so the skill is ahead of the published page.

lwshang (approved):

